PR #8, opened 2021-02-13
CMake: 3.17.3+

ax3l:
We don't use the DEVICE_LINK features in upstream AMReX logic yet, so we can stay with 3.17.3+ for now.

For CMake < 3.18.0 we need to overspecify the Python dependencies that we are actually looking for. This is mainly a problem with `manylinux` images (but there we can upgrade CMake with `pip` first) and fine-grained package managers.

Generally, anticipate a switch of AMReX as a whole to CMake 3.19+ in the mid-term because a lot of niceness landed in it.

Hand-crafted for @sayerhs so he does not need to upgrade his clusters on a Saturday :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 # Preamble ####################################################################
 #
-cmake_minimum_required(VERSION 3.18.0)
+cmake_minimum_required(VERSION 3.17.3)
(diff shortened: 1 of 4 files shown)


ax3l:
We decided to keep things clean and nice and update the systems :)

PR #10, opened 2021-02-13
setup.py: fix temporary build dirs

ax3l:
forgotten to update when I changed the structure

follow-up to #6

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -88,7 +88,7 @@ class CMakeBuild(build_ext):
 
         cmake_args = [
             '-DCMAKE_LIBRARY_OUTPUT_DIRECTORY=' +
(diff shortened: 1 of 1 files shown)


sayerhs:
@ax3l Instead of this change, may I recommend that we replace 

https://github.com/AMReX-Codes/pyamrex/blob/885cd9508d730653c868e6f1b4d29d8fe3c702e4/setup.py#L225

with 

```python
package_dir = {'': 'src'},
```

This will allow me to do `pip install -e .`

ax3l:
Yeah, also this change is invalid - it specifies where in the wheel we want to have the generated python module placed (and for shared amrex libs also their destination)

PR #14, opened 2021-02-14
Introduce AMReX configuration query object

sayerhs:
Use a Python class with read-only static properties to allow Python query of
AMReX build time configuration.

### DIFF
--- a/src/Base/AMReX.cpp
+++ b/src/Base/AMReX.cpp
@@ -11,6 +11,10 @@
 namespace py = pybind11;
 using namespace amrex;
 
+namespace {
+struct Config {};
+}
+
 void init_AMReX(py::module& m)
 {
     py::class_<AMReX>(m, "AMReX")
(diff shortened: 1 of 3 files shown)


src/Base/AMReX.cpp
```diff
@@ -21,6 +25,51 @@ void init_AMReX(py::module& m)
...
+            })
+        .def_property_readonly_static(
+            "have_gpu",
+            [](py::object){
+#ifdef AMREX_USE_GPU
+                return true;
+#else
+                return false;
+#endif
+            })
```
ax3l, line 57:
```suggestion
            })
        .def_property_readonly_static(
            "have_omp",
            [](py::object){
#ifdef AMREX_USE_OMP
                return true;
#else
                return false;
#endif
            })
```

ax3l (commented):
Nice, thank you!

src/Base/AMReX.cpp
```diff
@@ -21,6 +25,51 @@ void init_AMReX(py::module& m)
                     py::return_value_policy::reference)
         ;
 
+    py::class_<Config>(m, "Config")
```
ax3l, line 28: Shall we move this over to a `Config.cpp` file?

Can act as a py-counterpart to `AMReX_Config.H`.

sayerhs: I thought about that... but `Verbose(), OutStream(), Version()` are defined in `AMReX.H` so it felt better to keep it along with other `AMReX.H` functions.

sayerhs: May be we should name that class `Settings` instead if `Config` is confusing... that still doesn't play nice with `OutStream` and `Print` though :(

Also `SpaceDim` is in `AMReX_SPACE.H`

ax3l: Settings is a great idea, too.

I often called this `get(Build)Variant(s)` in the past, but that's probably not too obvious either.

ax3l: Slack: staying with *config* for now since its similar in other python packages (e.g. mpl)

ax3l (approved):

PR #17, opened 2021-02-15
[WIP] Array4: __array_interface__

### LINKED ISSUE #9: Discussion on mapping between amrex, numpy.ndarray, and torch.tensor data types
Hey, this is not so much an issue as a place to solicit public feedback.

I think we should implement type conversion from the amrex FArrayBox (or more precisely the Array4<double>) data type to numpy.ndarray and torch.tensor. As well a suitable python CUDA variants.

I also think that this type conversion should have s copying and a referencing variant.

This shouldn't be hard to implement (NO! This won't support python 2... I have a life you know), and I volunteer my time. But first I want to run this past all y'all to see if anyone is already working on it and what you think.

Tagging @ax3l @maxpkatz @drummerdoc

ax3l:
- [x] draft the `__array_interface__`
- [ ] handle C->F and F->C array index properly (invert shape & stride from/to AMReX)
- [ ] check how to handle offsets expressed with begin/end
- [ ] handle 1D-3D(4D) numpy array adoption properly
- [ ] write type dispatch wrapper for `Array4` construction
- [ ] implement `__cuda_array_interface__`
- [ ] :peace_symbol: :heart: :rainbow: 

Close #9

### DIFF
--- a/src/Base/Array4.cpp
+++ b/src/Base/Array4.cpp
@@ -1,0 +1,160 @@
+/* Copyright 2021 The AMReX Community
+ *
+ * Authors: Axel Huebl
+ * License: BSD-3-Clause-LBNL
+ */
+#include <pybind11/pybind11.h>
+#include <pybind11/numpy.h>
+#include <pybind11/stl.h>
+
+#include <AMReX_Config.H>
+#include <AMReX_Array4.H>
+#include <AMReX_IntVect.H>
+
+#include <sstream>
+#include <type_traits>
+
+namespace py = pybind11;
+using namespace amrex;
+
+
+template< typename T >
+void make_Array4(py::module &m, std::string typestr)
+{
+    // dispatch simpler via: py::format_descriptor<T>::format() naming
+    auto const array_name = std::string("Array4_").append(typestr);
+    py::class_< Array4<T> >(m, array_name.c_str(), py::buffer_protocol())
+        .def("__repr__",
+             [](Array4<T> const & a4) {
+                 std::stringstream s;
+                 s << a4.size();
+                 return "<amrex.Array4 of size '" + s.str() + "'>";
+             }
+        )
+#if defined(AMREX_DEBUG) || defined(AMREX_BOUND_CHECK)
+        .def("index_assert", &Array4<T>::index_assert)
+#endif
+
+        .def_property_readonly("size", &Array4<T>::size)
+        .def_property_readonly("nComp", &Array4<T>::nComp)
+
+        .def(py::init< >())
+        .def(py::init< Array4<T> const & >())
+        .def(py::init< Array4<T> const &, int >())
+        .def(py::init< Array4<T> const &, int, int >())
+        //.def(py::init< T*, Dim3 const &, Dim3 const &, int >())
+
+        .def(py::init([](py::array_t<T> & arr) {
+            py::buffer_info buf = arr.request();
+
+            auto a4 = std::make_unique< Array4<T> >();
+            a4.get()->p = (T*)buf.ptr;
+            a4.get()->begin = Dim3{0, 0, 0};
+            // TODO: likely C->F index conversion here
+            // p[(i-begin.x)+(j-begin.y)*jstride+(k-begin.z)*kstride+n*nstride];
+            a4.get()->end.x = (int)buf.shape.at(0);
+            a4.get()->end.y = (int)buf.shape.at(1);
+            a4.get()->end.z = (int)buf.shape.at(2);
+            a4.get()->ncomp = 1;
+            // buffer protocol strides are in bytes, AMReX strides are elements
+            a4.get()->jstride = (int)buf.strides.at(0) / sizeof(T);
+            a4.get()->kstride = (int)buf.strides.at(1) / sizeof(T);
+            a4.get()->nstride = (int)buf.strides.at(2) * (int)buf.shape.at(2) / sizeof(T);
+            return a4;
+        }))
+
+
+        .def_property_readonly("__array_interface__", [](Array4<T> const & a4) {
+            auto d = py::dict();
+            auto const len = length(a4);
+            // TODO: likely F->C index conversion here
+            // p[(i-begin.x)+(j-begin.y)*jstride+(k-begin.z)*kstride+n*nstride];
+            auto shape = py::make_tuple(  // Buffer dimensions
+                    len.x < 0 ? 0 : len.x,
+                    len.y < 0 ? 0 : len.y,
+                    len.z < 0 ? 0 : len.z//,  // zero-size shall not have negative dimension
+                    //a4.ncomp
+            );
+            // buffer protocol strides are in bytes, AMReX strides are elements
+            auto const strides = py::make_tuple(  // Strides (in bytes) for each index
+                    sizeof(T) * a4.jstride,
+                    sizeof(T) * a4.kstride,
+                    sizeof(T)//,
+                    //sizeof(T) * a4.nstride
+            );
+            d["data"] = py::make_tuple(long(a4.dataPtr()), false);
+            d["typestr"] = py::format_descriptor<T>::format();
+            d["shape"] = shape;
+            d["strides"] = strides;
+            // d["strides"] = py::none();
+            d["version"] = 3;
+            return d;
+        })
+
+        // not sure if useful to have this implemented on top
+/*
+        .def_buffer([](Array4<T> & a4) -> py::buffer_info {
+            auto const len = length(a4);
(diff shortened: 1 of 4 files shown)


src/Base/Array4.cpp
```diff
@@ -0,0 +1,160 @@
...
+void make_Array4(py::module &m, std::string typestr)
+{
+    // dispatch simpler via: py::format_descriptor<T>::format() naming
+    auto const array_name = std::string("Array4_").append(typestr);
+    py::class_< Array4<T> >(m, array_name.c_str(), py::buffer_protocol())
+        .def("__repr__",
+             [](Array4<T> const & a4) {
+                 std::stringstream s;
+                 s << a4.size();
+                 return "<amrex.Array4 of size '" + s.str() + "'>";
```
sayerhs, line 31: This should indicate the polymorphic type.

src/Base/Array4.cpp
```diff
@@ -0,0 +1,160 @@
...
+            auto d = py::dict();
+            auto const len = length(a4);
+            // TODO: likely F->C index conversion here
+            // p[(i-begin.x)+(j-begin.y)*jstride+(k-begin.z)*kstride+n*nstride];
+            auto shape = py::make_tuple(  // Buffer dimensions
+                    len.x < 0 ? 0 : len.x,
+                    len.y < 0 ? 0 : len.y,
+                    len.z < 0 ? 0 : len.z//,  // zero-size shall not have negative dimension
+                    //a4.ncomp
+            );
```
sayerhs, line 77: How does this handle ghost cells?

src/Base/Array4.cpp
```diff
@@ -0,0 +1,160 @@
...
+        .def(py::init< Array4<T> const &, int >())
+        .def(py::init< Array4<T> const &, int, int >())
+        //.def(py::init< T*, Dim3 const &, Dim3 const &, int >())
+
+        .def(py::init([](py::array_t<T> & arr) {
+            py::buffer_info buf = arr.request();
+
+            auto a4 = std::make_unique< Array4<T> >();
+            a4.get()->p = (T*)buf.ptr;
+            a4.get()->begin = Dim3{0, 0, 0};
```
sayerhs, line 52: This probably needs to be adjusted to a box index space

sayerhs (commented):
Looking good @ax3l seems like `numpy <-> Array4` roundtrip is working already?

Currently, I see a conflict between how a user would expect to index into an array in Python vs. AMReX C++ code. In AMReX we expect to index using a Box so the starting index will not be zero. And in presence of ghost cells, negative indices will yield the ghost cells on the `lo` side.

src/Base/Array4.cpp
```diff
@@ -0,0 +1,160 @@
...
+        .def(py::init< Array4<T> const & >())
+        .def(py::init< Array4<T> const &, int >())
+        .def(py::init< Array4<T> const &, int, int >())
+        //.def(py::init< T*, Dim3 const &, Dim3 const &, int >())
+
+        .def(py::init([](py::array_t<T> & arr) {
+            py::buffer_info buf = arr.request();
+
+            auto a4 = std::make_unique< Array4<T> >();
+            a4.get()->p = (T*)buf.ptr;
```
JBlaschke, line 51: If I read this correctly, this _aliases_ the `Array4` to the python buffer. Idk enough about the buffer protocol, so I am asking here: will `arr` be kept alive as long as this instance of `pyamrex.Array4`?

tests/test_array4.py
```diff
@@ -0,0 +1,86 @@
...
+    # from numpy (also a non-owning view)
+    x = np.ones((2, 3, 4,))
+    arr = amrex.Array4_double(x)
+
+    x[1, 1, 1] = 42
+    # TypeError: 'amrex.amrex_pybind.Array4_double' object is not subscriptable
+    # assert(arr[1, 1, 1] == 42)
+
+    # copy to numpy
+    c_arr2np = np.array(arr, copy=True)
```
JBlaschke, line 30: Dumb question -- but I need to learn eventually -- is this the "official" way to define a amrex -> python deepcopy (i.e. using the `np.array` constructor)? Or do we need to define a dedicated deepcopy also?

JBlaschke: Followup question: do we need to define a non-aliasing constructor (i.e. python -> amrex deepcopy)?

I can imaging that for data analytics applications we might want to keep a time series in python (looking at you machine learning folks) -- for that we need the former (amrex -> python) which is covered by this.

For experimental science applications we frequently have python provide a data source. A python -> amrex deepcopy (probably a simple copy constructor?) would be helpful in ensuring that the data is _owned_ by amrex.

Come to think of it: A lot of algorithms in amrex are involve creating copies of data and then applying functions to them. Users might want to preserve this workflow.

My opinion is that this case best approach would be to expose the Array4 copy constructor (assuming that hasn't been deleted), and settle on the workflow: i) shallow-copy into amrex; ii) call copy constructor to create a deep copy.

src/Base/Array4.cpp
JBlaschke: Also what happens when the user changes to contents of `arr`?

src/Base/Array4.cpp
ax3l: Not yet at all, let's have a VC about how you solved it in the past.

So far I am mainly experimenting with the data interface (trying the GPU version next)

src/Base/Array4.cpp
ax3l: Thanks, at the moment (and I am just trying the design of the data access) I made the Python Array4 behave like the C++ Array4: it's purely a view to data that is managed elsewhere.

If you create a numpy array and pass it to an Array4, the data is owned by Python. If you return an Array4 that is created on from C++ data, then the C++ side owns the data.

This might not be the final design because it can shoot Python users in the foot and I'll discuss with @sayerhs (feel free to join!) how to design this in the end. In his prior design, he basically did not expose Array4 as a user-creatable type at all but instead used it to return views to C++ data only. This sounds to me like something we might want to do again - but I need to catch up with him to see the details.

tests/test_array4.py
ax3l: I added some details above. This is just a view at the moment, the same way as the C++ Array4 is a plain view to non-owned data.

I just added the numpy overload so I can test around with it, maybe in the end we do not let users create this type directly because it will easily lead to memory issues if used wrongly (and segfaulting interpreters are never well received by users, even if they caused it by not following an API contract).

ax3l:
Discussed today:
- we will make Array4(Real/Int) createable only by MultiFabs (`amrex::Box` indexing semantics by default for everything: index `-1` is ghost)
- add explicit `as_array` / `as_cupy` functions for non-owning data from it: (`numpy`/Python Buffer Interface semantics for everything: index `-1` is end of array)
- make data copies explicit or fail if coming from device (check also pytorch & tensorflow interfaces for good vibes)

Refs.:
- https://github.com/sayerhs/pyamrex/blob/master/amrex/amrex_base.pyx#L502-L511
- https://github.com/sayerhs/pyamrex/blob/master/amrex/cpp/multifab.pxi
- https://sayerhs.github.io/pyamrex/index.html#amrex.amrex_base.MultiFab
```py
# Looping
for mfi in mfab:
    bx = mfi.tilebox()
    marr = mfab.array(mfi)

    for i, j, k in bx:
        mar[i, j, k, 0] = 10.0 * i
        mar[i, j, k, 1] = 10.0 * j
        mar[i, j, k, 2] = 10.0 * k
```

src/Base/Array4.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+#include <sstream>
+#include <type_traits>
+
+namespace py = pybind11;
+using namespace amrex;
+
+
+template< typename T >
+void make_Array4(py::module &m, std::string typestr)
```
ax3l, line 22: Discussed in #19: we want to make sure that users do not construct Array4s on their own and only get them from MFIter loops

ax3l:
This is now merged into #19 and improved therein.

src/Base/Array4.cpp
ax3l: Added in #19

src/Base/Array4.cpp
ax3l: Handled in #19

src/Base/Array4.cpp
ax3l: Note that this function here is mainly to support views into numpy and other python buffer objects via Array4.
In #19 currently mapped to a zero-based index space for incoming buffer protocols.

Potentially can implement shift-like functions later on to adjust the index space if we like to.

PR #30, opened 2022-03-26
Array4: __cuda_array_interface__ v3

ax3l:
Start implementing the `__cuda_array_interface__` for zero-copy data exchange on Nvidia CUDA GPUs.

Optional: accessing an external `__cuda_array_interface__` object in non-owning manner as AMReX Array4:
https://github.com/cupy/cupy/blob/a5b24f91d4d77fa03e6a4dd2ac954ff9a04e21f4/cupy/core/core.pyx#L2478-L2514

- [x] `mfab` and `mfab_device` need to become functions, not fixtures. Otherwise they will be cached and outlive `amrex.finalize()`: #81 #84
- [x] Particle Iter & MFIter: Python does not destruct keys by default: add a GPU stream synchronize here? https://github.com/AMReX-Codes/pyamrex/blob/78bbbc752ca3f65e257261385f1e31defb59f85e/src/Base/MultiFab.cpp#L72-L75
depends on https://github.com/AMReX-Codes/amrex/pull/2983 and https://github.com/AMReX-Codes/amrex/pull/2985
`if` and `for` do not create a scope in Python (they do in C++):
```py
In [1]: import numpy as np                                                                                                           

In [2]: x = np.array([1,2,3])                                                   

In [3]: for a in x: 
   ...:     print(a)                                                                
1
2
3

# a is still alive xD
In [4]: a                                                                       
Out[4]: 3
```

### DIFF
--- a/src/Base/Array4.cpp
+++ b/src/Base/Array4.cpp
@@ -140,6 +140,10 @@ void make_Array4(py::module &m, std::string typestr)
             return a4;
         }))
 
+        /* init from __cuda_array_interface__: non-owning view
+         * TODO
+         */
+
 
         // CPU: __array_interface__ v3
         // https://numpy.org/doc/stable/reference/arrays.interface.html
--- a/src/Base/Array4.cpp
+++ b/src/Base/Array4.cpp
@@ -164,9 +164,16 @@ void make_Array4(py::module &m, std::string typestr)
             // Because the user of the interface may or may not be in the same context, the most common case is to use cuPointerGetAttribute with CU_POINTER_ATTRIBUTE_DEVICE_POINTER in the CUDA driver API (or the equivalent CUDA Runtime API) to retrieve a device pointer that is usable in the currently active context.
             // TODO For zero-size arrays, use 0 here.
 
-            // ... TODO: wasn't there some stream or device info?
+            // None or integer
+            // An optional stream upon which synchronization must take place at the point of consumption, either by synchronizing on the stream or enqueuing operations on the data on the given stream. Integer values in this entry are as follows:
+            //   0: This is disallowed as it would be ambiguous between None and the default stream, and also between the legacy and per-thread default streams. Any use case where 0 might be given should either use None, 1, or 2 instead for clarity.
+            //   1: The legacy default stream.
+            //   2: The per-thread default stream.
+            //   Any other integer: a cudaStream_t represented as a Python integer.
+            //   When None, no synchronization is required.
+            d["stream"] = py::none();
 
-            d["version"] = 2;
+            d["version"] = 3;
             return d;
         })
 
--- a/src/Base/Array4.cpp
+++ b/src/Base/Array4.cpp
@@ -19,6 +19,61 @@ namespace py = pybind11;
 using namespace amrex;
 
 
+namespace
+{
+    /** CPU: __array_interface__ v3
+     *
+     * https://numpy.org/doc/stable/reference/arrays.interface.html
+     */
+    template<typename T>
+    py::dict
+    array_interface(Array4<T> const & a4)
+    {
+        auto d = py::dict();
+        auto const len = length(a4);
+        // F->C index conversion here
+        // p[(i-begin.x)+(j-begin.y)*jstride+(k-begin.z)*kstride+n*nstride];
+        // Buffer dimensions: zero-size shall not skip dimension
+        auto shape = py::make_tuple(
+                a4.ncomp,
+                len.z <= 0 ? 1 : len.z,
+                len.y <= 0 ? 1 : len.y,
+                len.x <= 0 ? 1 : len.x  // fastest varying index
+        );
+        // buffer protocol strides are in bytes, AMReX strides are elements
+        auto const strides = py::make_tuple(
+                sizeof(T) * a4.nstride,
+                sizeof(T) * a4.kstride,
+                sizeof(T) * a4.jstride,
+                sizeof(T)  // fastest varying index
+        );
+        bool const read_only = false;
+        d["data"] = py::make_tuple(std::intptr_t(a4.dataPtr()), read_only);
+        // note: if we want to keep the same global indexing with non-zero
+        //       box small_end as in AMReX, then we can explore playing with
+        //       this offset as well
+        //d["offset"] = 0;         // default
+        //d["mask"] = py::none();  // default
+
+        d["shape"] = shape;
+        // we could also set this after checking the strides are C-style contiguous:
+        //if (is_contiguous<T>(shape, strides))
+        //    d["strides"] = py::none();  // C-style contiguous
+        //else
+        d["strides"] = strides;
+
+        // type description
+        // for more complicated types, e.g., tuples/structs
+        //d["descr"] = ...;
+        // we currently only need this
+        d["typestr"] = py::format_descriptor<T>::format();
+
+        d["version"] = 3;
+        return d;
+    }
+}
+
+
 template< typename T >
 void make_Array4(py::module &m, std::string typestr)
 {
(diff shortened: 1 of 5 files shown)


ax3l:
:warning: There is an nvcc host code generation bug that we fixed with Nvidia last night. Affects CUDA Toolkit 11.4-11.8 with pybind11 (https://github.com/pybind/pybind11/issues/4193)
I will ship a work-around for pybind11 (https://github.com/pybind/pybind11/pull/4220) before the next CUDA release, pls use an older NVCC (e.g. 11.3) in the meantime.


Update: patch that unlocks that broken compiler range in https://github.com/pybind/pybind11/pull/4220
Add this to CMake:
```bash
cmake -S . -B build -DAMREX_GPU_BACKEND=CUDA -DpyAMReX_pybind11_repo=https://github.com/ax3l/pybind11.git -DpyAMReX_pybind11_branch=fix-nvcc-11.4-11.8
```

ax3l:
@RemiLehe build logic from README.md is this:

- https://github.com/AMReX-Codes/pyamrex#developers
- https://github.com/AMReX-Codes/pyamrex#build

So concretely:
```bash
# Python packages if not already installed as described
python3 -m pip install -U pip setuptools wheel
python3 -m pip install -U cmake pytest
python3 -m pip install -U -r requirements.txt

# depending on what you try
python3 -m pip install cupy-cuda11x
python3 -m pip install numba
python3 -m pip install torch

```
```bash
# configure once (unless changing backend or versions heavily)
cmake -S . -B build -DAMReX_GPU_BACKEND=CUDA \
    -DpyAMReX_pybind11_repo=https://github.com/ax3l/pybind11.git \
    -DpyAMReX_pybind11_branch=fix-nvcc-11.4-11.8
# rinse & repeat: builds, packages & runs pip install
cmake --build build --target pip_install -j 8
```

and tests:
```bash
# Run all tests
python3 -m pytest tests/

# Run tests from a single file
python3 -m pytest tests/test_array4.py

# Run a single test (useful during debugging)
python3 -m pytest tests/test_array4.py::test_array4_cupy
python3 -m pytest tests/test_multifab.py::test_mfab_ops_cuda_cupy

# Run all tests, do not capture "print" output and be verbose
python3 -m pytest -s -vvvv tests/test_array4.py
```

and with nsight:
```
nsys profile -f true -t cuda,nvtx,osrt python3 -m pytest -s -vvv tests/test_multifab.py::test_mfab_ops_cuda_cupy
```
GUI:
```
nsight-sys
```

ax3l:
Found a tiny bug, will rebase after #77 was merged. - Update: done.

Found another arena bug, will rebase after #78 was merged. - Update: done.

ax3l:
Screenshot from 2022-10-06 20-24-00
First cupy progress. Gotta learn how to do in-place updates on arrays in kernels...

tests/test_multifab.py
```diff
@@ -152,3 +152,132 @@ def test_mfab_mfiter(mfab, nghost):
...
+            # write and read into the marr_cupy
+            marr_cupy[()] = 0.0
+            set_to_seven(marr_cupy)
+
+    # verify
+    with cupy.prof.time_range("verify 7", color_id=3):
+        sum = mfab_device.sum_unique(comp=0, local=False)
+        assert sum == shape * 7
+
+    # TODO: @jit.rawkernel()
```
ax3l, line 280: Could try to explicitly `del mfab_device` here to see if that fixes the memory crash in `~FabArray` during test shutdown.

ax3l: Or the fixtures for classes such as `mfab`/`mfab_device` should be adjusted to use `function` scope:
https://docs.pytest.org/en/stable/reference/fixtures.html#fixture-instantiation-order

ax3l: This should fix it: #78

ax3l: rebased

tests/test_multifab.py
```diff
@@ -152,3 +152,132 @@ def test_mfab_mfiter(mfab, nghost):
...
+    with cupy.prof.time_range("verify 5", color_id=2):
+        sum = mfab_device.sum_unique(comp=0, local=False)
+        assert sum == shape * 5
+
+    # assign 7
+    with cupy.prof.time_range("assign 7 (fuse)", color_id=3):
+
+        @cp.fuse(kernel_name="set_to_seven")
+        def set_to_seven(x):
+            x += 7.0
```
ax3l: Ideas to get rid of the extra `cudaMemset` before the kernel:
```suggestion
            x *= 0
            x += 7.0
```

ax3l:
```suggestion
            x[()] = 7.0 + x
```

ax3l:
With the new `MFIter::Finalize`, I can also see the `cudaStreamSynchronize` calls at the end of the iteration :)
Screenshot from 2022-10-13 18-00-34
Screenshot from 2022-10-13 18-01-38

tests/test_multifab.py
ax3l:
```suggestion
            x[...] = 7.0
```

tests/test_multifab.py
```diff
@@ -152,3 +154,161 @@ def test_mfab_mfiter(mfab, nghost):
...
+
+        for mfi in mfab_device:
+            bx = mfi.tilebox().grow(ngv)
+            marr = mfab_device.array(mfi)
+            marr_cupy = cp.array(marr, copy=False)
+
+            # write and read into the marr_cupy
+            marr_cupy[()] = 0.0
```
ax3l:
```suggestion
```

src/Base/Array4.cpp
```diff
@@ -85,56 +140,44 @@ void make_Array4(py::module &m, std::string typestr)
...
+        .def_property_readonly("__cuda_array_interface__", [](Array4<T> const & a4) {
+            auto d = array_interface(a4);
+
+            // data:
+            // Because the user of the interface may or may not be in the same context, the most common case is to use cuPointerGetAttribute with CU_POINTER_ATTRIBUTE_DEVICE_POINTER in the CUDA driver API (or the equivalent CUDA Runtime API) to retrieve a device pointer that is usable in the currently active context.
+            // TODO For zero-size arrays, use 0 here.
+
+            // ... TODO: wasn't there some stream or device info?
+
+            d["version"] = 2;
```
ax3l: Actually at v3 now:
https://numba.readthedocs.io/en/latest/cuda/cuda_array_interface.html
```suggestion
        // Nvidia GPUs: __cuda_array_interface__ v2
        // https://numba.readthedocs.io/en/latest/cuda/cuda_array_interface.html
        .def_property_readonly("__cuda_array_interface__", [](Array4<T> const & a4) {
            auto d = array_interface(a4);

            // data:
            // Because the user of the interface may or may not be in the same context, the most common case is to use cuPointerGetAttribute with CU_POINTER_ATTRIBUTE_DEVICE_POINTER in the CUDA driver API (or the equivalent CUDA Runtime API) to retrieve a device pointer that is usable in the currently active context.
            // TODO For zero-size arrays, use 0 here.

            // None or integer
            // An optional stream upon which synchronization must take place at the point of consumption, either by synchronizing on the stream or enqueuing operations on the data on the given stream. Integer values in this entry are as follows:
            //   0: This is disallowed as it would be ambiguous between None and the default stream, and also between the legacy and per-thread default streams. Any use case where 0 might be given should either use None, 1, or 2 instead for clarity.
            //   1: The legacy default stream.
            //   2: The per-thread default stream.
            //   Any other integer: a cudaStream_t represented as a Python integer.
            //   When None, no synchronization is required.
            d["stream"] = py::none();

            d["version"] = 3;
```

ax3l:
Wuup, wuup. First part done.
Larger tests and particles next :)

PR #38, opened 2022-05-10
AMReX: Turn particles on

ax3l:
Particles are by default turned off in AMReX. This turns them on in pyAMReX.

### DIFF
--- a/cmake/dependencies/AMReX.cmake
+++ b/cmake/dependencies/AMReX.cmake
@@ -25,8 +25,7 @@ macro(find_amrex)
         set(AMReX_FORTRAN OFF CACHE INTERNAL "")
         set(AMReX_FORTRAN_INTERFACES OFF CACHE INTERNAL "")
         set(AMReX_BUILD_TUTORIALS OFF CACHE INTERNAL "")
-        # potentially worth changing default:
-        #set(AMReX_PARTICLES ON CACHE INTERNAL "")
+        set(AMReX_PARTICLES ON CACHE INTERNAL "")  # default: OFF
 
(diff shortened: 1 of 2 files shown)


cmake/dependencies/AMReX.cmake
```diff
@@ -66,7 +65,7 @@ macro(find_amrex)
     else()
         message(STATUS "Searching for pre-installed AMReX ...")
         # https://amrex-codes.github.io/amrex/docs_html/BuildingAMReX.html#importing-amrex-into-your-cmake-project
-        find_package(AMReX 21.02 CONFIG REQUIRED COMPONENTS PIC)
+        find_package(AMReX 21.02 CONFIG REQUIRED COMPONENTS PARTICLES PIC)
```
RTSandberg, line 68: What does this addition do?

RTSandberg (approved):
I'm ready for AMReX_PARTICLES :)

cmake/dependencies/AMReX.cmake
ax3l: We can find AMReX in three ways:
- download & compile it (default) or
- pass a source location and compile it or
- find an already built (compiled) and installed AMReX library.

This option here is the last case. We look for a pre-compiled AMReX build and we make sure that it was compiled with particles enabled.

References:
- https://amrex-codes.github.io/amrex/docs_html/BuildingAMReX.html#building-with-cmake
- https://cmake.org/cmake/help/latest/command/find_package.html

RTSandberg: Great, thanks!

PR #39, opened 2022-05-10
add length to Box, fix RealBox init

RTSandberg:
- Add length() to Box
- Add a test function to test_box.py
- Fix RealBox initialization with 6 reals (not two lists of 3 reals each)
- Fix corresponding test


ax3l (approved):
Thanks :)

PR #42, opened 2022-05-17
CMake Pip: Unique Custom Targets

ax3l:
Add a prefix to custom targets so we do not collide if used as a CMake subproject, e.g., in ImpactX or WarpX.


RTSandberg (approved):
This makes sense with the intended use of pyamrex in other projects

ax3l:
Restarting CI (#41 merged)

PR #44, opened 2022-06-13
CI: Fix Windows Clang

ax3l:
Fix:
```
CMake Error at C:/hostedtoolcache/windows/Python/3.9.13/x64/Lib/site-packages/cmake/data/share/cmake-3.22/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Python (missing: Python_LIBRARIES Python_INCLUDE_DIRS
  Interpreter Development.Module)
```

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -33,6 +33,9 @@ jobs:
     if: github.event.pull_request.draft == false
     steps:
     - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
     - uses: seanmiddleditch/gha-setup-ninja@master
     - name: Build & Install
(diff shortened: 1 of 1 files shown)


.github/workflows/windows.yml
```diff
@@ -42,14 +45,10 @@ jobs:
         cmake -S . -B build               `
               -T "ClangCl"                `
               -DCMAKE_VERBOSE_MAKEFILE=ON `
-              -DAMReX_MPI=OFF             `
-              -DPython_EXECUTABLE=C:\\hostedtoolcache\\windows\\python\\3.9.12\\x64\\python3.exe
```
ax3l, line 46: Mainly because that version changed to `3.9.13` - but generalized the logic now

PR #48, opened 2022-06-15
Tests: Increase AMReX Verbosity

ax3l:
Keep AMReX status output on for tests.


RTSandberg (approved):
This is a good idea for the tests - maybe a little noisy but the alternative is silent ninja CI death

ax3l:
Sounds good. If too noisy, we can also go for `amrex.verbose=1`, which removes some of the additional statistic prints of level 2.

PR #50, opened 2022-07-19
Fix: MultiFab Iterator (first)

ax3l:
Python iterates by calling initially:
- `__iter__`
- `__next__`

This means we need to handle the first iterator element explicitly, otherwise we will jump directly to the 2nd element. We do this now the same way as pybind11 does this, via a little state:
  https://github.com/pybind/pybind11/blob/v2.10.0/include/pybind11/pybind11.h#L2269-L2282

### DIFF
--- a/src/Base/MultiFab.cpp
+++ b/src/Base/MultiFab.cpp
@@ -37,7 +37,7 @@ void init_MultiFab(py::module &m) {
 
     py::class_< FArrayBox >(m, "FArrayBox");
 
-    py::class_< MFIter >(m, "MFIter")
+    py::class_< MFIter >(m, "MFIter", py::dynamic_attr())
         .def("__repr__",
              [](MFIter const & mfi) {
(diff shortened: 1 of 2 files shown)


src/Base/MultiFab.cpp
```diff
@@ -59,16 +59,23 @@ void init_MultiFab(py::module &m) {
         //     [](MFIter & mfi) -> MFIter & {
         //        return mfi;
         //     },
-        //     py::return_value_policy::reference
+        //     py::return_value_policy::reference_internal
         //)
         .def("__next__",
              [](MFIter & mfi) -> MFIter & {
-                ++mfi;
+                static bool first_or_done = true;
```
sayerhs: Can you create multiple iterators?

ax3l: I was thinking about this as well, I should make this a state per object

Issue #49, opened 2022-07-14
Particle declarations to header file for generality and to avoid downstream collisions


RTSandberg:
Move specific particle declarations and creation to separate header file to allow applications to use the pyAMReX library and declare their own particle sizes.  Consider creating a testing submodule with the particle initializations currently declared.

PR #51, opened 2022-07-22
Particles: More Specializations

ax3l:
Add a few more template specializations as used in ImpactX and HiPACE++ particle containers.

### DIFF
--- a/src/Particle/ArrayOfStructs.cpp
+++ b/src/Particle/ArrayOfStructs.cpp
@@ -80,7 +80,8 @@ void make_ArrayOfStructs(py::module &m)
 }
 
 void init_ArrayOfStructs(py::module& m) {
-    make_ArrayOfStructs<7,0> (m);
+    make_ArrayOfStructs< 0, 0> (m);
+    make_ArrayOfStructs< 7, 0> (m);
     make_ArrayOfStructs< 1, 1> (m);
     make_ArrayOfStructs< 2, 1> (m);
 }
--- a/src/Particle/Particle.cpp
+++ b/src/Particle/Particle.cpp
@@ -321,8 +321,11 @@ void init_Particle(py::module& m) {
     ;
     py::enum_<PIdx::IntValues>(pidx, "IntValues")
     ;
-    make_Particle<PIdx::nRealAttribs, PIdx::nIntAttribs> (m);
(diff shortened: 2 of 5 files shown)


src/Particle/ArrayOfStructs.cpp
```diff
@@ -80,7 +80,8 @@ void make_ArrayOfStructs(py::module &m)
 }
 
 void init_ArrayOfStructs(py::module& m) {
-    make_ArrayOfStructs<7,0> (m);
+    make_ArrayOfStructs< 0, 0> (m);
```
ax3l, line 83: @RTSandberg this one should cause no problems, does it?
```C++
     make_ArrayOfStructs< 0, 0> (m);
```

RTSandberg: I can't think of any reason it should cause problems, we even have some tests in place to keep the user from accidentally referencing non-existent real or int data

src/Particle/Particle.cpp
```diff
@@ -321,8 +321,11 @@ void init_Particle(py::module& m) {
...
-    make_Particle<1, 1> (m);
-    make_Particle<2, 1> (m);
-    make_Particle<3, 2> (m);
+    make_Particle< PIdx::nRealAttribs, PIdx::nIntAttribs > (m);
+    make_Particle< 1, 1 > (m);
+    make_Particle< 2, 1 > (m);
+    make_Particle< 3, 2 > (m);
+    make_Particle< 4, 0 > (m);  // HiPACE++ 22.07
+    make_Particle< 5, 0 > (m);  // ImpactX 22.07
+    make_Particle< 37, 1> (m);  // HiPACE++ 22.07
```
RTSandberg, line 330: Thanks for the comments indicating which versions are for which codes

RTSandberg (commented):
Looks good to me

src/Particle/ArrayOfStructs.cpp
ax3l: Awesome, thanks! :)

src/Particle/Particle.cpp
ax3l: Sure thing, thanks :)

Thank you for documenting the task in #49 as well :sparkles:

PR #52, opened 2022-07-25
add GeometryData and test

RTSandberg:
Add `GeometryData` class, test of its functionality, and the `Geometry::data()` function which returns a `GeometryData` object


ax3l (approved):
Awesome, thanks a lot :rocket: :sparkles:

PR #53, opened 2022-07-27
add name to namespace for Config

RTSandberg:
This is a fix for a bug that was observed when loading `pyImpactX`.  The `amrex` and `impactx` `Config` objects were both described in anonymous namespaces causing a conflict.  Here the namespace is explicitly named so that the `Config` object is `amrex.Config`.


ax3l (approved):
Thank you! :)

PR #62, opened 2022-08-07
Add More `Box` Methods

ax3l:
Add more methods to `Box`.

- [x] I think `hi_vect`/`lo_vect` are equal to `small_end`/`big_end` and need to be reimplemented to return not an `int*` - which results in returning a scalar of the first value - but an array
- [x] We should implement that `IntVect` can be either casted to a list or be iterated

## To Do

- [x] I don't understand why the `test_surrounding_nodes` does not pass though.

### DIFF
--- a/tests/test_box.py
+++ b/tests/test_box.py
@@ -6,8 +6,7 @@ import amrex
 
 @pytest.fixture
 def box():
-    #return amrex.Box((0, 0, 0), (127, 127, 127))
-    return amrex.Box(amrex.IntVect(0, 0, 0), amrex.IntVect(127, 127, 127))
+    return amrex.Box((0, 0, 0), (127, 127, 127))
 
 def test_length(box):
     print(box.length())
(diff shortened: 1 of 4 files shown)


ax3l:
I don't understand why the `test_surrounding_nodes` does not pass though.

tests/test_box.py
```diff
@@ -46,12 +47,14 @@ def test_length(box):
...
 @pytest.mark.parametrize("dir", [-1, 0, 1, 2])
 def test_surrounding_nodes(box, dir):
     """Surrounding nodes"""
-    nx = np.array(box.hi_vect)
+    nx = box.hi_vect
+    print('nx=', nx)
     bx = box.surrounding_nodes(dir=dir)
+    print('bx=', bx)
 
     if dir < 0:
```
ax3l: Issues below this line for the `test_surrounding_nodes` test.
Maybe we create a copy here and things go south... the `surrounding_nodes` method seems to manipulate the object itself and only returns a reference to itself (for convenience/method chaining?) again.

ax3l: Looks like there was some extra logic added in Cython for this test to work:
https://github.com/sayerhs/pyamrex/blob/f2909f90e0c43e45829a7275a77ef009fd726c40/amrex/amrex_base.pyx#L106-L128

sayerhs: @ax3l cython doesn't have a clean way to support overloaded functions. So `-1` was used as a proxy for the surrounding nodes method without any arguments. You can probably cleanup the logic in pybind11

ax3l: Oh, thanks a lot for the hint - that's the info I needed :)

sayerhs (approved):

PR #70, opened 2022-09-07
CMake Python Tests: Multi-Config Hints

ax3l:
Make sure that we can find the extra build config suffixx directory that is appended in multi-config generators.

Re: https://github.com/ECP-WarpX/impactx/pull/208


CMakeLists.txt
```diff
@@ -133,6 +133,28 @@ set_target_properties(pyAMReX PROPERTIES
     PDB_OUTPUT_DIRECTORY ${CMAKE_PYTHON_OUTPUT_DIRECTORY}/amrex
     COMPILE_PDB_OUTPUT_DIRECTORY ${CMAKE_PYTHON_OUTPUT_DIRECTORY}/amrex
 )
+if(GENERATOR_IS_MULTI_CONFIG)
```
ax3l:
```suggestion
get_property(isMultiConfig GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
if(isMultiConfig)
```

PR #75, opened 2022-09-19
CMake Defaults: No Cache

ax3l:
Setting this w/o cache limits the scope to the current project and does not influence superprojects that do not set the same variables.

See:
- https://github.com/openPMD/openPMD-api/pull/1313
- and issue description in https://github.com/openPMD/openPMD-api/pull/1312

### DIFF
--- a/cmake/pyAMReXFunctions.cmake
+++ b/cmake/pyAMReXFunctions.cmake
@@ -46,25 +46,17 @@ endmacro()
 #
 macro(pyamrex_set_default_build_dirs)
     if(NOT CMAKE_ARCHIVE_OUTPUT_DIRECTORY)
-        set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib"
-                CACHE PATH "Build directory for archives")
-        mark_as_advanced(CMAKE_ARCHIVE_OUTPUT_DIRECTORY)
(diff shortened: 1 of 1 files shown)


cmake/pyAMReXFunctions.cmake
```diff
@@ -77,11 +72,9 @@ macro(pyamrex_set_default_install_dirs)
...
         include(GNUInstallDirs)
         if(NOT CMAKE_INSTALL_CMAKEDIR)
-            set(CMAKE_INSTALL_CMAKEDIR "${CMAKE_INSTALL_LIBDIR}/cmake"
-                CACHE PATH "CMake config package location for installed targets")
+            set(CMAKE_INSTALL_CMAKEDIR "${CMAKE_INSTALL_LIBDIR}/cmake")
             if(WIN32)
-                set(CMAKE_INSTALL_LIBDIR Lib
-                    CACHE PATH "Object code libraries")
+                set(CMAKE_INSTALL_LIBDIR Lib)
                 set_property(CACHE CMAKE_INSTALL_CMAKEDIR PROPERTY VALUE "cmake")
```
RTSandberg:
```suggestion
```
It looks like this line is failing in the Windows CI; is the problem that ``Cache`` no longer has ``CMAKE_INSTALL_CMAKEDIR`` as a variable (this being the purpose of the PR) ?

ax3l: Thanks, fixed now! :)

PR #87, opened 2022-10-20
pyAMReX: Bind Particle Iterators

ax3l:
Implement particle (box) iterators.

- [x] tests

### DIFF
--- a/tests/test_particleContainer.py
+++ b/tests/test_particleContainer.py
@@ -121,6 +121,58 @@ def test_pc_init():
     assert pc.TotalNumberOfParticles() == pc.NumberOfParticlesAtLevel(0) == npart
     assert pc.OK()
 
+    print("Iterate particle boxes & set values")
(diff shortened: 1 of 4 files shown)


tests/test_particleContainer.py
```diff
@@ -121,6 +121,58 @@ def test_pc_init():
...
+        aos_arr[0]["y"] = 0.35
+        aos_arr[0]["z"] = 0.40
+
+        # TODO: this seems to write into a copy of the data
+        soa = pti.soa()
+        real_arrays = soa.GetRealData()
+        int_arrays = soa.GetIntData()
+        real_arrays[0] = [0.55]
+        real_arrays[1] = [0.22]
+        int_arrays[0] = [2]
```
ax3l, line 147: @RTSandberg I think the SoA logic does not yet work to write and creates a copy here...?

PR #88, opened 2022-10-21
Helpers: to_numpy/cupy

### LINKED ISSUE #55: .to_numpy(), .to_cupy(), etc.
For all objects that expose the `__array_interface__`, we should add a helper member function called `.to_numpy()` that does nothing but create a view:
```py
np.array(self, copy=False, order='F')
cupy.array(self, copy=False, order='F')
```

Similar to:
- https://github.com/openPMD/openPMD-api/blob/b05171d371fe9568149ef58ae60ee113f4d0c4b4/examples/11_particle_dataframe.py#L34
- https://github.com/openPMD/openPMD-api/blob/b05171d371fe9568149ef58ae60ee113f4d0c4b4/examples/11_particle_dataframe.py#L49
- https://github.com/openPMD/openPMD-api/blob/b05171d371fe9568149ef58ae60ee113f4d0c4b4/examples/11_particle_dataframe.py#L87

Equivalently, we want to add `.to_copy()` et al. functions for #30 and later DLPack interfaces.

ax3l:
- [x] Array4
- [x] PODVector_real/int (aka RealVector/IntVector)
- [x] SoA
- [x] AoS
- [ ] ParticleTile
- [ ] ParticleContainer (list of non-owning objects)
- [x] MultiFab (list of non-owning objects)

Close #55

### DIFF
--- a/src/amrex/Array4.py
+++ b/src/amrex/Array4.py
@@ -1,0 +1,98 @@
+"""
+This file is part of pyAMReX
+
+Copyright 2023 AMReX community
+Authors: Axel Huebl
+License: BSD-3-Clause-LBNL
+"""
+
+
+def array4_to_numpy(self, copy=False, order="F"):
+    """
+    Provide a Numpy view into an Array4.
+
+    Note on the order of indices:
+    By default, this is as in AMReX in Fortran contiguous order, indexing as
+    x,y,z. This has performance implications for use in external libraries such
+    as cupy.
+    The order="C" option will index as z,y,x and perform better with cupy.
+    https://github.com/AMReX-Codes/pyamrex/issues/55#issuecomment-1579610074
+
+    Parameters
+    ----------
+    self : amrex.Array4_*
+        An Array4 class in pyAMReX
+    copy : bool, optional
+        Copy the data if true, otherwise create a view (default).
+    order : string, optional
+        F order (default) or C. C is faster with external libraries.
+
+    Returns
+    -------
+    np.array
+        A numpy n-dimensional array.
+    """
+    import numpy as np
+
+    if order == "F":
+        return np.array(self, copy=copy).T
+    elif order == "C":
+        return np.array(self, copy=copy)
+    else:
+        raise ValueError("The order argument must be F or C.")
+
+
+def array4_to_cupy(self, copy=False, order="F"):
+    """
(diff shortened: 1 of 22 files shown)


src/amrex/Array4.py
```diff
@@ -0,0 +1,98 @@
...
+
+    Returns
+    -------
+    np.array
+        A numpy n-dimensional array.
+    """
+    import numpy as np
+
+    if order == "F":
+        return np.array(self, copy=copy).T
```
ax3l, line 38: Follow-up: support AMReX-side device-to-host copies.

ax3l, line 38: Follow-up: support AMReX-side device-to-host copies.

ax3l: Added directly to MultiFab in #192

ax3l: Added directly to MultiFab in #192

src/amrex/StructOfArrays.py
```diff
@@ -0,0 +1,99 @@
...
+    soa_view = SoA_np([], [])
+
+    if self.size() == 0:
+        raise ValueError("SoA is empty.")
+
+    for idx_real in range(self.NumRealComps()):
+        soa_view.real.append(self.GetRealData(idx_real).to_numpy(copy=copy))
+
+    for idx_int in range(self.NumIntComps()):
+        soa_view.int.append(self.GetIntData(idx_int).to_numpy(copy=copy))
```
ax3l, line 41: Double check if runtime attributes are part of this:
https://github.com/AMReX-Codes/pyamrex/pull/220#discussion_r1382548333

src/amrex/StructOfArrays.py
```diff
@@ -0,0 +1,99 @@
...
+    soa_view = SoA_cp([], [])
+
+    if self.size() == 0:
+        raise ValueError("SoA is empty.")
+
+    for idx_real in range(self.NumRealComps()):
+        soa_view.real.append(self.GetRealData(idx_real).to_cupy(copy=copy))
+
+    for idx_int in range(self.NumIntComps()):
+        soa_view.int.append(self.GetIntData(idx_int).to_cupy(copy=copy))
```
ax3l, line 81: See above: Double check if runtime attributes are part of this.

PR #91, opened 2022-11-04
Support Shared AMReX Lib

ax3l:
- [x] CI: Add a Linux and Windows build that each build a shared AMReX library.
- [x] fix issues: For Windows Python 3.8+, this shared `libamrex.dll` needs to be injected via `PATH` and `os.add_dll_directory`
  - https://github.com/python/cpython/issues/80266
  - https://docs.python.org/3.8/library/os.html#os.add_dll_directory

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -62,11 +62,11 @@ jobs:
 
         llvm-objdump -p  "C:\Program Files (x86)\pyAMReX\bin\amrex.dll"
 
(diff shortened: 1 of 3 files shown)


.github/workflows/windows.yml
```diff
@@ -46,13 +46,20 @@ jobs:
...
               -DCMAKE_VERBOSE_MAKEFILE=ON `
+              -DBUILD_SHARED_LIBS=ON      `
               -DAMReX_MPI=OFF
         if(!$?) { Exit $LASTEXITCODE }
 
         cmake --build build --config Debug -j 2
         if(!$?) { Exit $LASTEXITCODE }
-        cmake --build build --config Debug --target pip_install -j 2
+
+        dumpbin /dependents build\lib\site-packages\amrex\amrex_pybind.cp311-win_amd64.pyd
```
```suggestion
        llvm-objdump -p build\lib\site-packages\amrex\amrex_pybind.cp311-win_amd64.pyd
```

PR #119, opened 2023-04-08
ParticleTileData Interface Changes

ax3l:
This keeps the legacy AoS-position-id particle layout in AMReX.

It just anticipates the changes in C++ APIs for `ArrayOfStruct` and `(Const)ParticleTileData` from https://github.com/AMReX-Codes/amrex/pull/2878

- [x] solve geometry periodicity change issue in test first #120
- [x] undo CMake change (pre-PR merge)
- [x] do NOT merge before upstream PR

### DIFF
--- a/src/Particle/ParticleTile.cpp
+++ b/src/Particle/ParticleTile.cpp
@@ -14,6 +14,7 @@
 
 #include <sstream>
 
+
 namespace py = pybind11;
 using namespace amrex;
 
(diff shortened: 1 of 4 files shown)


src/Particle/ParticleTile.cpp
```diff
@@ -66,17 +76,19 @@ void make_ParticleTile(py::module &m, std::string allocstr)
...
-        .def("size", &ParticleTileType::size)
-        .def("numParticles", &ParticleTileType::numParticles)
-        .def("numRealParticles", &ParticleTileType::numRealParticles)
-        .def("numNeighborParticles", &ParticleTileType::numNeighborParticles)
-        .def("numTotalParticles", &ParticleTileType::numTotalParticles)
+        //.def("size", &ParticleTileType::size) // TODO
+        //.def("numParticles", &ParticleTileType::numParticles) // TODO
+        //.def("numRealParticles", &ParticleTileType::numRealParticles) // TODO
+        //.def("numNeighborParticles", &ParticleTileType::numNeighborParticles) // TODO
+        //.def("numTotalParticles", &ParticleTileType::numTotalParticles) // TODO
```
ax3l: To check: Did we accidentally drop these or are they now at a different type?

src/Particle/ParticleTile.cpp
```diff
@@ -35,21 +40,26 @@ void make_ParticleTileData(py::module &m) {
             .def_readonly("m_size", &ParticleTileDataType::m_size)
             .def_readonly("m_num_runtime_real", &ParticleTileDataType::m_num_runtime_real)
             .def_readonly("m_num_runtime_int", &ParticleTileDataType::m_num_runtime_int)
-            .def("getSuperParticle", &ParticleTileDataType::getSuperParticle)
+            //.def("getSuperParticle", &ParticleTileDataType::getSuperParticle) // TODO
```
ax3l: Unclear where this went, need to check.

src/Particle/ParticleTile.cpp
ax3l: Ah, maybe need to be explicit to resolve the sfinae in the methods.

src/Particle/ParticleTile.cpp
ax3l: also: now an SFINAE. Need to add template to help the compiler to resolve it.

PR #122, opened 2023-04-12
[WIP] Tests w/o managed Memory

ax3l:
TODO: We need to update the tests so they do not need to rely on managed memory.


ax3l:
Already addressed in other PRs by now and merged with the option in this PR.

PR #127, opened 2023-05-04
[Experiment] Nanobind

ax3l:
What would it take to migrate to nanobind?

## Pro's

See advertising, all great :)

Love the
- speed & size
- separate headers for features
- non-header-only approach
- C++17 and newer only
- faster builds: no LTO needed anymore

## Notes / Con's

- limited official compiler support (Clang, GCC, MSVC), especially no coverage yet for GPU (Nvidia NVCC & NVHPC, AMD ROCm, Intel OneAPI)
- still many warnings (e.g., Wshadow, Wpedantic) in base lib
- ~currently, limited public docs~ improved :)


BenWibking:
@ax3l I am very interested to know whether nanobind is capable of handling pyAMReX in 2026.

This would make adding pyAMReX support to Quokka a lot easier due to reduced compile times.

ax3l:
I agree. Most tricky part is compiler coverage and single maintainer approach in nanobind, which means we have to fork for every workaround we face in HPC.

Nonetheless, 3 years after this PR, LLMs make trying this fast. Let's do a fresh PR with one.

I am on PTO a few days but back mid of the week. Feel free to take a stab if you have a few token left ;)

BenWibking:
Sure, I'll create a vibe-coded test PR :)

BenWibking:
See https://github.com/AMReX-Codes/pyamrex/pull/595.

PR #131, opened 2023-05-25
Fix: CMake-Built Wheel Install

ax3l:
- [x] test install from CMake works on Unix/Windows
- [x] fix CMake-triggered wheels build (pre-built library pickup)

Follow-up to #129

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -66,5 +66,6 @@ jobs:
         if(!$?) { Exit $LASTEXITCODE }
     - name: Unit tests as Installed
       run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/windows.yml
```diff
@@ -64,3 +64,7 @@ jobs:
         if(!$?) { Exit $LASTEXITCODE }
         cmake --build build --config Debug --target pip_install
         if(!$?) { Exit $LASTEXITCODE }
+    - name: Unit tests as Installed
+      run: |
+        rm -rf build
```
```suggestion
        rm -r -fo build
```

PR #140, opened 2023-06-20
Add more BoxArray features

dpgrote:


ax3l (approved):
Splendid!

Just realized we do not have pytest standalone tests for this class in `tests/` yet. If you like to add some as a follow-up that would be appreciated :)

PR #145, opened 2023-07-13
wrapper for concatenate function in amrex

RevathiJambunathan:
In this PR, we add a wrapper to the concatenate function in amrex.

Thanks to the team effort by @EZoni @oshapoval @kngott and Scot Halverson.

### DIFF
--- a/src/Base/Utility.cpp
+++ b/src/Base/Utility.cpp
@@ -12,10 +12,8 @@ using namespace amrex;
 void init_Utility(py::module& m)
 {
     m.def("concatenate",
-          [] (const std::string& root,
-                     int                num,
-                     int                mindigits) {
-          return amrex::Concatenate(root, num, mindigits);
-          }, py::return_value_policy::move,
-          "Builds plotfile name");
+          &amrex::Concatenate,
+          "Builds plotfile name",
+          py::arg("root"), py::arg("num"), py::arg("mindigits")=5
+    );
 }
--- a/src/Base/Utility.cpp
+++ b/src/Base/Utility.cpp
@@ -1,6 +1,7 @@
 /* Copyright 2021-2022 The AMReX Community
  *
  * License: BSD-3-Clause-LBNL
+ * Authors: Revathi Jambunathan, Axel Huebl
  */
 #include <AMReX_Utility.H>
 #include <pybind11/pybind11.h>
--- a/src/Base/Utility.cpp
+++ b/src/Base/Utility.cpp
@@ -1,0 +1,21 @@
+/* Copyright 2021-2022 The AMReX Community
+ *
+ * License: BSD-3-Clause-LBNL
+ */
+#include <AMReX_Utility.H>
+#include <pybind11/pybind11.h>
+#include <sstream>
+
+namespace py = pybind11;
+using namespace amrex;
+
+void init_Utility(py::module& m)
+{
+    m.def("concatenate",
+          [] (const std::string& root,
+                     int                num,
+                     int                mindigits) {
+          return amrex::Concatenate(root, num, mindigits);
+          }, py::return_value_policy::move,
+          "Builds plotfile name");
+}
--- a/src/pyAMReX.cpp
+++ b/src/pyAMReX.cpp
@@ -82,8 +82,8 @@ PYBIND11_MODULE(amrex_3d_pybind, m) {
                Periodicity
                PODVector
                StructOfArrays
-               Vector
                Utility
+               Vector
     )pbdoc";
 
     // note: order from parent to child classes
--- a/src/pyAMReX.cpp
+++ b/src/pyAMReX.cpp
@@ -43,7 +43,7 @@ void init_ParticleContainer(py::module &);
 void init_Periodicity(py::module &);
 void init_PODVector(py::module &);
 void init_Vector(py::module &);
-
+void init_Utility(py::module &);
 
 #if AMREX_SPACEDIM == 1
 PYBIND11_MODULE(amrex_1d_pybind, m) {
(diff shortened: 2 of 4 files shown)


EZoni:
@RevathiJambunathan 
Discussed offline with @ax3l, besides using the wrapper when converting the AMReX tutorials to pyamrex, it would be useful to add a small test of this function within one of the existing tests in the pyamrex repository.

src/pyAMReX.cpp
```diff
@@ -114,6 +115,9 @@ PYBIND11_MODULE(amrex_3d_pybind, m) {
     init_ParticleContainer(m);
     init_AmrMesh(m);
 
+    // Wrappers around functions, independent of parent to child order
```
ax3l: In fact, these are dependent on the order of types if an argument or return type defined above is used :)

Here, in fact, no AMReX type is (yet) used, but other functions we might expose in the future in `AMReX_Utility.H` do use AMReX types that we need to register first, like `amrex::Vector`.

src/pyAMReX.cpp
```diff
@@ -83,6 +83,7 @@ PYBIND11_MODULE(amrex_3d_pybind, m) {
                PODVector
                StructOfArrays
                Vector
+               Utility
```
ax3l: Keeping alphabetic order :)
```suggestion
               Utility
               Vector
```

src/Base/Utility.cpp
```diff
@@ -0,0 +1,21 @@
...
+
+void init_Utility(py::module& m)
+{
+    m.def("concatenate",
+          [] (const std::string& root,
+                     int                num,
+                     int                mindigits) {
+          return amrex::Concatenate(root, num, mindigits);
+          }, py::return_value_policy::move,
+          "Builds plotfile name");
```
ax3l: Since `amrex::Concatenate` is very simple in terms of arguments and has no overloads, we can simply use its function pointer.

Then, we can use the default return policy, copying the string out to Python by value.

Lastly, let's name the user-facing function arguments with `py::arg`:
```suggestion
          &amrex::Concatenate,
          "Builds plotfile name",
          py::arg("root"), py::arg("num"), py::arg("mindigits")=5
    );
```

src/pyAMReX.cpp
```diff
@@ -43,7 +43,7 @@ void init_ParticleContainer(py::module &);
 void init_Periodicity(py::module &);
 void init_PODVector(py::module &);
 void init_Vector(py::module &);
-
+void init_Utility(py::module &);
```
ax3l: This could be alphabetical for clarity as well, went a bit astray in recent PRs :sweat_smile:

RevathiJambunathan: yeah -- we were going to keep it alphabetical but saw that it was not really that way 

Then we discussed that may be we can keep a distinction between amrex classes and functions and left a space there

RevathiJambunathan: what would be the best practice here, keeping all class wrappers together i.e., separate from functions wrappers -- or merge them , but keep it alphabetical

RevathiJambunathan: For example, somewhere in that file, RealBox comes before Before Array 
not alphabetical

ax3l: Yes, there slipped in some other inconsistencies here that I will fix separately.

For docs and declarations, let's keep them alphabetical for easy reading.
On the call site, we need to be careful to call init routines of dependent classes and functions that use AMReX types after we declared them.

src/pyAMReX.cpp
RevathiJambunathan: ah so should I remove the comment?

ax3l: Maybe this?
```suggestion
    // Wrappers around standalone functions
```

src/pyAMReX.cpp
```diff
@@ -42,9 +42,9 @@ void init_ParticleTile(py::module &);
 void init_ParticleContainer(py::module &);
 void init_Periodicity(py::module &);
 void init_PODVector(py::module &);
+void init_Utility(py::module &);
 void init_Vector(py::module &);
```
ax3l, line 47: Readding newline :)
```suggestion


```

tests/test_utility.py
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
+
+import numpy as np
+import pytest
```
ax3l:
```suggestion
```

tests/test_utility.py
```diff
@@ -0,0 +1,11 @@
+# -*- coding: utf-8 -*-
+
+
+import amrex.space3d as amr
+
+
+def test_concatenate():
+    pltname = amr.concatenate("plt", 1000, 5)
+    print("--test concatenate --")
+    print("plotfile name", pltname)
```
ax3l:
```suggestion
```

tests/test_utility.py
```diff
@@ -0,0 +1,9 @@
+# -*- coding: utf-8 -*-
+
+
```
ax3l:
```suggestion
```

ax3l (approved):
This looks great, thank you! :rocket: :sparkles:

src/Base/Utility.cpp
```diff
@@ -0,0 +1,19 @@
+/* Copyright 2021-2022 The AMReX Community
+ *
+ * License: BSD-3-Clause-LBNL
```
ax3l, line 3:
```suggestion
 * License: BSD-3-Clause-LBNL
 * Authors: Revathi Jambunathan, Axel Huebl
```

PR #146, opened 2023-07-13
Add to ParallelDescriptor.cpp

dpgrote:
Adds a few more routines to ParallelDescriptor.cpp, the easy routines without arguments.


ax3l:
Intel CI fix coming via #147

ax3l (approved):
Thank you! :)

PR #149, opened 2023-07-19
Include pyAMReX.H & ODR Fix

ax3l:
We provide a single helper include now for all pyAMReX compilation units (TUs) / cpp files.

This ensures that optional pybind11 includes, e.g., for stl, numpy, functions and operators, are consistently present in all TUs. This is required to avoid C++ one-definition-rule (ODR) violations, which are undefined behavior and thus could cause instabilities down the road.


RTSandberg (approved):
This looks good to me!

dpgrote (approved):
This looks good to me!

PR #156, opened 2023-07-21
Add initialized and size to AMReX

dpgrote:
Add handy routines to AMReX


src/Base/AMReX.cpp
```diff
@@ -101,6 +101,9 @@ void init_AMReX(py::module& m)
           }, py::return_value_policy::reference,
           "Initialize AMReX library");
 
+    m.def("initialized", &Initialized);
+    m.def("size", &AMReX::size);
```
ax3l: Handy! Can you please add their doc strings?

ax3l (approved):
Thank you :)

Issue #76, opened 2022-09-28
Handle AMReX Abort Across Dynamic Libs to Apps


ax3l:
We want to make sure that calls to `amrex::Abort()` & general exceptions from, e.g., `throw std::runtime_error(...)`s, can be propagated across dynamic libraries of AMReX, ABLASTR, etc. up to app Python bindings. In other words: an `amrex::Abort()` called from routines that are compiled into the AMReX `.so` file, e.g. `amrex::ParmParse::get()`, should be able to be caught as a runtime error in WarpX/ImpactX.

See description in https://github.com/ECP-WarpX/impactx/pull/265 for details.

- [ ] set `amrex.signal_handling = 0` and `amrex.throw_exception = 1` in Python runs
- [ ] proper `amrex::Abort` support: enable working, cross-dynamic-library `throw std::runtime_error` exceptions
  - [ ] macOS: e.g., RTTI flags on AppleClang ?
  - [ ] Linux: e.g., `-shared-libgcc` on Linux (or use `g++` for linking)
  - [ ] since RTTI is the default with all modern compilers, this could also be related to symbol stripping in pybind11 release builds: https://stackoverflow.com/a/17923224/2719194 - which conda-forge seems to handle well (see links below)
- [ ] make sure users can still trigger backtraces if they want to

Might also addressable via a custom `system::error_handler`:
https://github.com/AMReX-Codes/amrex/blob/b84d7c069cef7470f195b250926ca0e84ec46fb2/Src/Base/AMReX.cpp#L205-L206

Refs.:
- https://stackoverflow.com/questions/38878999/try-catch-doesnt-work-in-shared-library/38879990#38879990
- https://stackoverflow.com/questions/3638237/polymorphically-catching-an-exception-in-a-fno-rtti-shared-library-on-mac-os-x
- On conda-forge, [run-time type information](https://en.wikipedia.org/wiki/Run-time_type_information) (RTTI) is the default (and used for more stuff than exceptions, e.g., `dynamic_cast`, `typeid` over library bounds)
- https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html?highlight=rtti
- conda-forge:
  - https://github.com/conda-forge/libcxx-testing-feedstock/blob/7b267de95fc212c0c4235f76315314c1d4d5231d/recipe/meta.yaml#L26-L36
  - https://github.com/search?q=org%3Aconda-forge+rtti&type=code

ax3l:
X-ref:
[...]

PR #159, opened 2023-07-22
Fill out min and max

dpgrote:
This enhances the min and max wrappers, adding keyword arguments with defaults, error checking of the input, and doc strings.

It would be nice to have this level of wrapping for all of the routines, but this would be a fair amount of work. The error checking is fairly important since wrong values can give either just bad results or seg faults.

### DIFF
--- a/src/Base/MultiFab.cpp
+++ b/src/Base/MultiFab.cpp
@@ -17,6 +17,18 @@
 #include <memory>
 #include <string>
 
+namespace {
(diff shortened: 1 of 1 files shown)


ax3l (approved):
Thank you, this is great!

Some of the error checking, e.g., if used outside of hot loops, could also be done on the AMReX C++ side.

A clean `throw` will propagate to our level as long as we compile with RTTI enabled #76

PR #174, opened 2023-08-11
Add amrex::Print() Like Function

ax3l:
Add `amr.Print(...)` that behaves like Python's `print()`, but only prints on the AMReX IOProcessor.

- [x] add a test

### DIFF
--- a/src/amrex/space1d/__init__.py
+++ b/src/amrex/space1d/__init__.py
@@ -32,3 +32,9 @@ __author__ = amrex_1d_pybind.__author__
 #
 def d_decl(x, y, z):
     return (x,)
+
+
+def Print(*args, **kwargs):
+    """Wrap amrex::Print() - only the IO processor writes"""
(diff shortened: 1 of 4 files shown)


src/amrex/space1d/__init__.py
```diff
@@ -32,3 +32,9 @@
 #
 def d_decl(x, y, z):
     return (x,)
+
+
+def Print(*args, **kwargs):
+    """Wrap amrex::Print() - only the IO processor writes"""
+    if ParallelDescriptor.IOProcessor():
```
ax3l: These bindings to `ParallelProcessor` segfault if they are called before AMReX initialize.

I wonder if we should either in AMReX or at least in pyAMReX add a test for `amrex::initialized` and then do something cleaner (throw or all print).

ax3l: pushed an update

PR #181, opened 2023-08-28
pytest: Temporary Working Directory

ax3l:
Tests can generate temporary output and directories, which might clash and/or need cleaning between tests.

This changes the current working directory to a unique directory per test.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 import itertools
+import os
 
 import pytest
 
@@ -17,6 +18,9 @@ except ImportError:
 
 if amr.Config.have_mpi:
     from mpi4py import MPI
+
+# base path for input files
+basepath = os.getcwd()
 
 
 @pytest.fixture(autouse=True, scope="function")
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -20,20 +20,21 @@ if amr.Config.have_mpi:
 
 
 @pytest.fixture(autouse=True, scope="function")
-def amrex_init():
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -20,20 +20,21 @@
...
+                # AMReX backtrace files: allows to attach to
+                # debuggers
+                "amrex.throw_exception=1",
+                "amrex.signal_handling=0",
+                # abort GPU runs if out-of-memory instead of swapping to host RAM
+                # "abort_on_out_of_gpu_memory=1",
+            ]
+        )
     yield
     amr.finalize()
```
sayerhs: Recommend that these two lines are also inside the `with` block

sayerhs (approved):

tests/conftest.py
ax3l: I was googling about `with` context and `yield` combinations and got nervous about what I saw :sweat_smile:

ax3l: Bravely trying my best:
```suggestion
        yield
        amr.finalize()
```

ax3l:
Hm, this actually fails my tests in https://github.com/ECP-WarpX/impactx/pull/424 ... maybe because it works now (and cannot find input files)?

sayerhs:
> Hm, this actually fails my tests in [ECP-WarpX/impactx#424](https://github.com/ECP-WarpX/impactx/pull/424) ... maybe because it works now (and cannot find input files)?

What is the error? How are you creating the input files?

ax3l:
Ah, now managed to pass their path before going into the per-test directory :)

PR #188, opened 2023-09-07
GitHub cache and ccache

WeiqunZhang:
Use ccache in CIs. Set up GitHub Cache for storing the ccache data.

TODO:
- [ ] Ccache does not work on Windows. Maybe we need to provide hints to cmake #190
- [x] ~~Ubuntu / GNU@7.5 has 60% cache miss rate for unknown reasons.~~ The reason seems to be the cache size was too small.
- [x] ~~The CleanUpCache workflow does not work. It is supposed to triggered by the the completion of "👑 CI". But either the space (between the crown and the word CI) or the emoji is causing issues.~~ The reason is this PR has not been merged yet. So the default branch does not have the workflow. (It would be a security issue if it updates cache according to a PR's instruction.)
- [x] Does CleanUpCachePostPR work? No reason it will not.
- [x] CodeQL

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,3 +33,18 @@ jobs:
     name: 🔄 Update Stub Files
     needs: [ubuntu, intel, hip, macos, windows]
     uses: ./.github/workflows/stubs.yml
+
+  save_pr_number:
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Save PR number
+        env:
+          PR_NUMBER: ${{ github.event.number }}
+        run: |
+          echo $PR_NUMBER > pr_number.txt
+      - uses: actions/upload-artifact@v3
+        with:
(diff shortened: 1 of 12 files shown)


WeiqunZhang:
I think everything except Windows works now as expected.

ax3l:
The reason that CCache on Windows does not work might be that the VS build script generator is not supporting ccache (yet)?
https://stackoverflow.com/a/73188028/2719194
https://cmake.org/cmake/help/v3.27/prop_tgt/LANG_COMPILER_LAUNCHER.html

I do that to make sure multi-config build generators work as expected, too.

We use Ninja as generator on AMReX Windows CI.
We can use Ninja for the Win Clang Debug build, too.

WeiqunZhang:
Unfortunately, the Ninja generator does not support toolset. I have restored windows.yml, since it does not work.

ax3l:
Thank you, let's address Windows CCache in a separate PR.

PR #196, opened 2023-09-29
Copy to Host: Array4.to_numpy(copy=True)

ax3l:
Support `to_numpy` copies w/o shared memory from device memory views in `Array4`.

X-ref #55


src/Base/Array4.H
```diff
@@ -0,0 +1,279 @@
...
+
+                // allocate host memory copy
+                auto h_data = py::array_t<std::remove_cv_t<T>>(shape, strides);
+
+                // sync copy: host data is unpinned
+                Gpu::copy(Gpu::deviceToHost,
+                    a4.dataPtr(), a4.dataPtr() + a4.size(),
+                    h_data.mutable_data()
+                );
+                return h_data;
```
ax3l: @WeiqunZhang does this look correct to you?

WeiqunZhang: Yes

ax3l: Thx. I am still debugging a copy issue that confuses me, the moment I have guard cells something goes off...

WeiqunZhang: At the BaseFab/Array4 level, there is no concept of ghost cells in the BaseFab class.

ax3l: Ah, my strides are still in bytes and need to be in elements for `py::array_t`.
Update: nope, needs to be bytes.

WeiqunZhang: I see. But is shape in elements? If so, that seems inconsistent.

ax3l: Oops, no... I forgot to set my desired value in all components of the multi fab

PR #208, opened 2023-10-18
CI: Add `ruff`

ax3l:
This adds [ruff](https://github.com/astral-sh/ruff) for Python code linting in pre-commits.

~~This is performed via `flake8 --select=F`, skipping all other stylistic checks of flake8 (which we perform with other tools such as black and isort). Also, `flake8` supports `...  # noqa` annotations to [silence warnings](https://github.com/PyCQA/pyflakes/issues/431) from `pyflakes`.~~

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -86,17 +86,6 @@ repos:
   - id: isort
     name: isort (python)
 
-# Python: Flake8 (checks only, does this support auto-fixes?)
-#- repo: https://github.com/PyCQA/flake8
-#  rev: 4.0.1
-#  hooks:
-#  - id: flake8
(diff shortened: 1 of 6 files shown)


tests/test_periodicity.py
```diff
@@ -1,36 +1,35 @@
...
+    assert not obj.is_all_periodic
     assert obj.is_periodic(0)
-    assert obj.is_periodic(1) == False
+    assert not obj.is_periodic(1)
     assert obj.is_periodic(2)
-    assert obj.is_periodic[2]
+    assert obj[2]
 
     bx = obj.domain
     print(bx)
```
ax3l: This line seems to crash. Curious.

Issue #207, opened 2023-10-18
Windows: PC Issue on 23.10


ax3l:
First seen in https://github.com/conda-forge/warpx-feedstock/pull/72 for only two our of many windows builds.

## win_64_numpy1.26python3.12.____cpython
```
2023-10-18T01:03:28.0503494Z (%PREFIX%) %SRC_DIR%>%PREFIX%\python.exe Examples\Physics_applications\laser_acceleration\PICMI_inputs_3d.py 
2023-10-18T01:03:28.5345324Z Traceback (most recent call last):
2023-10-18T01:03:28.5347941Z   File "D:\bld\warpx_1697583183195\_test_env\Lib\site-packages\pywarpx\_libwarpx.py", line 109, in load_library
2023-10-18T01:03:28.5349533Z     from . import warpx_pybind_3d as cxx_3d
2023-10-18T01:03:28.5351616Z ImportError: generic_type: type "PinnedMemoryParticleContainer" referenced unknown base type "amrex::ParticleContainer_impl<amrex::Particle<0,0>,4,0,amrex::PinnedArenaAllocator,amrex::DefaultAssignor>"
```

## win win_64_numpy1.22python3.9.____73_pypy
```
[...]

PR #209, opened 2023-10-19
Fix: Pinned Allocators First

ax3l:
They are used in methods that copy from device to host and thus need to be known first, before other allocator flavors.

This should help with the first issue in #207

Introduced in #196 #192 #88 #55


ax3l:
CUDA local runtime tests fail:
- /home/axel/src/pyamrex/tests/test_aos.py", line 29 in test_aos_push_pop

PR #212, opened 2023-10-20
Test: Periodicity (Enable 3D)

ax3l:
Fix ill-written `skipif` condition for pytest.


ax3l:
Some issue in `print(bx)`...

Issue #81, opened 2022-10-15
AMReX Initialize/Finalize as Context Manager


ax3l:
We cannot rely on the last `del`/reference drop to an object to really release and free the object in Python:
  https://stackoverflow.com/a/17644536/2719194

```py
import amrex

amrex.initialize([])
mfab = amrex.MultiFab( ... )
del mfab  # freed in CPython, but not guaranteed by Python per se
amrex.finalize()
```

This code can lead to Python holding on to `mfab` over the lifetime of `amrex.finalize()`. When using memory in an AMReX area, their actual `free` after `finalize()` will be UB and usually segfault.

Refs.:
- https://stackoverflow.com/a/17644536/2719194
- https://doc.pypy.org/en/latest/gc_info.html

WeiqunZhang:
The Arenas in amrex is fine in this.  That's why BArena never dies.

ax3l:
Yes, but tests that run on pure device MultiFabs should be tearing down cleanly as well before `finalize()`.
```py
import amrex

amrex.initialize([])
mfab = amrex.MultiFab(
    boxarr,
    distmap,
    num_components,
    num_ghost,
    amrex.MFInfo().set_arena(amrex.The_Device_Arena()),
)
del mfab  # freed in CPython, but not guaranteed by Python per se
amrex.finalize()
```

WeiqunZhang:
The device arena is also fine.
```
Arena*
The_Device_Arena ()
{   
    if        (the_device_arena) {
        return the_device_arena;
    } else {
        return The_Null_Arena();
    }
}
```

WeiqunZhang:
`The_Null_Arena` also never dies.  The device memory has been freed at that point.  As long as the memory in MultiFab is not being used, it will be fine.

ax3l:
Hm, I get a double free after finalize on process shutdown when I keep such an `mfab` alive #30 (update: maybe that is just an unsafe pointer/lease that we handed out, DLPack might make this safer)

ax3l:
Also, I kinda want to be sure that I have a context where all Arena (CPU/GPU) memory is free'd again. In testing, we need to run 100s of small init/finalize tests in the same process quickly after reach other - they should not leak memory after each of of these tests.

ax3l:
Oh I see what you mean, finalize will free all memory we allocated in arenas? @WeiqunZhang

WeiqunZhang:
[...]

PR #214, opened 2023-10-21
MultiFab Fixture Cleanup via `FabArray::clear`

ax3l:
Using a context manager and calling clear ensures that we will not hold device memory anymore once we hit `AMReX::Finalize`, even in the situation where an exception is raised in a test. This avoids segfaults for failing tests.

Related to #81


tests/conftest.py
```diff
@@ -85,47 +85,59 @@ def distmap(boxarr):
...
+            self.mfab.set_val(0.0, 0, num_components)
+            return self.mfab
 
-    return create
+        def __exit__(self, exc_type, exc_value, traceback):
+            self.mfab.clear()
+            del self.mfab
+
+    with MfabContextManager() as mfab:
+        yield mfab
```
sayerhs, line 109: Could be simplified?

```python
def mfab(boxarr, distmap, request):
    num_components = request.param[0]
    num_ghost = request.param[1]
    try:
        mfab = amr.MultiFab(boxarr, distmap, num_components, num_ghost)
        mfab.set_val(0.0, 0, num_components)
        yield mfab
    finally:
        mfab.clear()
        del mfab

File: tests/conftest.py [hunk 2]
```diff
@@ -85,47 +85,59 @@ def distmap(boxarr):
...
-        num_components = request.param[0]
-        num_ghost = request.param[1]
-        mfab = amr.MultiFab(boxarr, distmap, num_components, num_ghost)
-        mfab.set_val(0.0, 0, num_components)
-        return mfab
+    class MfabContextManager:
+        def __enter__(self):
+            num_components = request.param[0]
+            num_ghost = request.param[1]
+            self.mfab = amr.MultiFab(boxarr, distmap, num_components, num_ghost)
```
sayerhs, line 100: `pylint` would probably create warnings for this.

tests/conftest.py
ax3l: According to my knowledge, that will cache the MultiFab between runs and break out memory pool assumptions...

Worth to try again

tests/conftest.py
ax3l: Oh really? Linting still coming in #208 😅

CodeQL did not yet complain, what's to be avoided here?

tests/conftest.py
sayerhs: Interesting. The fixture is marked function scope, so it shouldn't cache a value. Is it pytest that's caching the value?

tests/conftest.py
sayerhs: Pylint will complain about attributes being declared/created in a non-init function.

tests/conftest.py
ax3l: I thought so... maybe I misunderstood something when caching kicks in.

ax3l:
@sayerhs I wonder if I shall just merge this for now and refactor it later, since it fixes issues we see on Perlmutter...?

PR #220, opened 2023-11-05
Pandas: `ParticleContainer_*.to_df()`

ax3l:
**Copy** all particles into a `pandas.DataFrame`. Supports local and MPI-gathered results.

## Future Directions (Attribute Names in AMReX)

At the moment, we make up generic names for the AoS and SoA attributes when we convert them to numpy, cupy and pandas.

It would generally be nice if AMReX supported the functionality to name particle attributes. In WarpX, we added this as `NamedParticleContainer`, but as a mainstream implementation we could avoid deriving yet another class type (because if we do, we will again have additional compile-time to bind that custom type in pyAMReX), and instead could just extend the particle classes with these runtime details.

- https://github.com/AMReX-Codes/amrex/issues/3614

## Future Directions (Write)

Once the AMReX SoA transition is complete in:
- https://github.com/AMReX-Codes/amrex/pull/3585
- https://github.com/AMReX-Codes/amrex/pull/3280

we might even be able to provide node-local, zero-copy **write access** in `Pandas` and `cuDF` particle data frames... (follow-up PRs).

### DIFF
--- a/src/amrex/ParticleContainer.py
+++ b/src/amrex/ParticleContainer.py
@@ -1,0 +1,104 @@
+"""
+This file is part of pyAMReX
+
+Copyright 2023 AMReX community
+Authors: Axel Huebl
+License: BSD-3-Clause-LBNL
+"""
+
+
+def pc_to_df(self, local=True, comm=None, root_rank=0):
+    """
+    Copy all particles into a pandas.DataFrame
+
+    Parameters
+    ----------
+    self : amrex.ParticleContainer_*
+        A ParticleContainer class in pyAMReX
+    local : bool
+        MPI-local particles
+    comm : MPI Communicator
+        if local is False, this defaults to mpi4py.MPI.COMM_WORLD
+    root_rank : MPI root rank to gather to
+        if local is False, this defaults to 0
+
+    Returns
+    -------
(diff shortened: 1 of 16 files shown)


src/amrex/ParticleContainer.py
```diff
@@ -0,0 +1,104 @@
...
+            # SoA
+            soa_view = pti.soa().to_numpy(copy=True)
+            soa_np_real = soa_view.real
+            soa_np_int = soa_view.int
+
+            for idx, array in enumerate(soa_np_real):
+                next_df[f"SoA_real_{idx}"] = array
+            for idx, array in enumerate(soa_np_int):
+                next_df[f"SoA_int_{idx}"] = array
+
```
ax3l, line 60: todo: I think this might miss on the runtime attributes (they are also in SoA format).

This should go into `SoA::to_numpy` aka `soa_to_numpy` & `soa_to_cupy`,

ax3l: Requires to call:
- `AddRealComp`
- query current number of real components
- call `ResizeRuntimeRealComp` again with redundant info (again the comm and number of components)

but https://github.com/AMReX-Codes/amrex/pull/3615  could simplify this.

PR #259, opened 2024-02-06
CI: 4 Cores Linux/Win, 3 Cores macOS

ax3l:
Increase build and test parallelism according to new increased core limits on [public GH hosted runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories).

image

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -17,6 +17,8 @@ jobs:
         python-version: '3.x'
     - name: Build & Install
       run: |
+        export CMAKE_BUILD_PARALLEL_LEVEL=4
+
         python3 -m pip install -U pip
         python3 -m pip install -U pandas pytest
         python3 -m pip install -v .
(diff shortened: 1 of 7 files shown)


WeiqunZhang:
Could also update cuda.yml and clang-tidy.yml. And there are a few places in intel, hip and ubuntu that are still using 2 processes.

WeiqunZhang:
and one more in windows.

ax3l:
I can't find `cuda.yml` and `clang-tidy.yml`. Wrong repo?

ax3l:
Found a few more! :)
(`pip install -v .` parallelism can be hinted via an env var)

WeiqunZhang:
> I can't find `cuda.yml` and `clang-tidy.yml`. Wrong repo?

Oh right, I thought this was WarpX repo.

WeiqunZhang (approved):

.github/workflows/windows.yml
```diff
@@ -17,6 +17,8 @@ jobs:
         python-version: '3.x'
     - name: Build & Install
       run: |
+        export CMAKE_BUILD_PARALLEL_LEVEL=4
```
```suggestion
        set "CMAKE_BUILD_PARALLEL_LEVEL=4"
```

PR #271, opened 2024-02-28
Python bindings for AMReX-MPMD 

siddanib:
This PR is a **draft** that includes python bindings for Multiple Program Multiple Data (MPMD) feature of AMReX. This feature depends on MPI for the involved C++ executable and mpi4py for the involved python script. **The submitted python bindings are based on the latest AMReX_MPMD.H and AMReX_MPMD.cpp**.


.gitignore
```diff
@@ -34,6 +34,24 @@
...
+
+## All compiled programs
+**/*.gnu.*.ex
+**/*.o
+**/*.ex
+
+## MacOS creates dynamic linker symbol files
+*.dSym
+
+## MacOS crap
```
ax3l: :sweat_smile: 
```suggestion
## MacOS custom folder attributes
```

CMakeLists.txt
```diff
@@ -115,6 +115,15 @@ foreach(D IN LISTS AMReX_SPACEDIM)
         $<BUILD_INTERFACE:${pyAMReX_SOURCE_DIR}/src>
     )
 
+    if(${AMReX_MPI})
+      execute_process(COMMAND
+        ${Python_EXECUTABLE} -c "import mpi4py; print( mpi4py.get_include() )"
```
ax3l: As a note, this does not work when cross-compiling. I will work on an alternative that does not needs mpi4py to be present and executable at compile time.

- https://github.com/openPMD/openPMD-api/blob/c2e0e83f0380fad0a7996cb6a496d0f66d2d5b6b/src/binding/python/Series.cpp#L38-L52
- https://github.com/openPMD/openPMD-api/blob/c2e0e83f0380fad0a7996cb6a496d0f66d2d5b6b/src/binding/python/Series.cpp#L226-L308

.gitignore
```diff
@@ -34,6 +34,24 @@
 *.out
 *.app
 
+## Compiler debris
```
ax3l: No big deal here, but for pull requests please try to address one aspect at a time. Changes to this file would be best done in a separate PR :)

.gitignore
siddanib: My bad :sweat:

CMakeLists.txt
siddanib: Thanks, Axel :+1: ! I included the dependence on mpi4py as my project involves a downstream task that depends on mpi4py. Actually getting just the C++ version of MPI_Comm from https://github.com/AMReX-Codes/amrex/blob/development/Src/Base/AMReX_MPMD.cpp#L97 and directly passing it back to amrex::Initialize worked in my initial attempts.

.gitignore
ax3l: No worries :D

CMakeLists.txt
ax3l: Hi @siddanib,

Yes the approach I linked will work with `mpi4py` :)
It just changes the build process to build without having to provide a run-able `mpi4y` at configuration and compilation time, which is needed when cross-compiling for different architectures.

.gitignore
ax3l: Moved to #280

ax3l:
@siddanib I pushed the update I suggested in review. Can you give it a try if this still works for you? :)

ax3l (approved):

src/Base/MPMD.cpp
```diff
@@ -0,0 +1,171 @@
...
+              const bool build_parm_parse = (cargs.size() > 1);
+
+              //! TODO perform mpi4py import test and check min-version
+              //!       careful: double MPI_Init risk? only import mpi4py.MPI?
+              //!       required C-API init? probably just checks:
+              //! refs:
+              //! -
+              //! https://bitbucket.org/mpi4py/mpi4py/src/3.0.0/demo/wrap-c/helloworld.c
+              //! - installed: include/mpi4py/mpi4py.MPI_api.h
+              // if( import_mpi4py() < 0 ) { here be dragons }
```
ax3l:
```suggestion
              auto m_mpi4py = py::module::import("mpi4py");
              amrex::ignore_unused(m_mpi4py);
```

ax3l:
From Slack:

Could you please consider this particular branch for testing the previous approach (https://github.com/siddanib/pyamrex/tree/development).
Could you please use this particular case for testing (https://github.com/siddanib/pyamrex/tree/development/tests/test_MPMD/MultiFab_two_way). `AMREX_HOME` in GNUmakefile needs to be changed.
I installed pyAMReX  as follows:
```
cmake -S . -B build -DAMReX_SPACEDIM="1;2;3" -DpyAMReX_amrex_src=/home/bsiddani/siddanib_github/amrex
cmake --build build -j 8 --target pip_install
```
For the test case:
I first build the `main.cpp` file as follows
```
make -j 10
```
Then I run the MPMD  case as follows
```
mpirun -np 8 ./main3d.gnu.DEBUG.MPI.ex : -np 4 python main.py
```
I will submit two clean examples once we figure out this issue. Please let me know if any additional information is needed. Thanks!

tests/test_MPMD/test_1/main.py
```diff
@@ -8,15 +8,9 @@
 # License: BSD-3-Clause-LBNL
 # Authors: Bhargav Sriram Siddani, Revathi Jambunathan, Edoardo Zoni, Olga Shapoval, David Grote, Axel Huebl
 
-import mpi4py
```
siddanib: The following lines:
```
import mpi4py
mpi4py.rc.initialize = False
mpi4py.rc.finalize = False
```
were resulting in `MPI initialized with thread support level 0` instead of `MPI initialized with thread support level 3`.

Removing these lines results in `MPI initialized with thread support level 3` and avoids the following error:
```
Abort(3755781) on node 10 (rank 10 in comm 0): Fatal error in internal_Comm_free: Invalid communicator, error stack:
internal_Comm_free(87): MPI_Comm_free(comm=0x7f03e8f576b0) failed
internal_Comm_free(44): Null communicator
```
Need to look into this further.

ax3l: Slack: solved :tada:

ax3l (approved):
Thank you, LGTM! :)

As a follow-up PR, can you add a `tests/test_MPMD/test_1/README.rst` file that describes the exact steps how to run this test?

ax3l:
I have to unbreak the macOS CI separately :)
Just a breaking change in homebrew/pip interaction.

PR #276, opened 2024-03-08
Particles: Restructure

ax3l:
The AMReX particle logic is heavily templated. For simpler maintenance, we move all declarations into header file and define needed types recursively from the particle container.

### DIFF
--- a/src/Particle/Particle.H
+++ b/src/Particle/Particle.H
@@ -3,6 +3,8 @@
  * Authors: Ryan Sandberg
  * License: BSD-3-Clause-LBNL
  */
+#pragma once
+
 #include "pyAMReX.H"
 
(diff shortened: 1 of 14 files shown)


src/Particle/Particle.H
```diff
@@ -50,6 +38,17 @@ namespace
 template <int T_NReal, int T_NInt=0>
 void make_Particle(py::module &m)
 {
+    // For legacy AoS + SoA particles, we register a particle and superparticle.
+    // The latter adds the SoA attributes. Avoid to double-register those types.
+    /*
+    static std::unordered_set<std::array<int, 2>> registered;
```
ax3l, line 44: Not needed yet, for later.

PR #282, opened 2024-03-25
Add `.copy` for MultiFab

RemiLehe:
Many thanks to @ax3l for writing most of the code here.


ax3l (approved):
:tada:

ax3l:
If AMReX merges https://github.com/AMReX-Codes/amrex/pull/3848, then we can simplify this to a one-liner :)

PR #287, opened 2024-04-03
Doc: MultiFab Args & Doc Strings

ax3l:
Add comprehensive doc strings to MultiFab.

Document that `to_numpy`/`to_cupy` include guard cells.

### DIFF
--- a/src/Base/MultiFab.cpp
+++ b/src/Base/MultiFab.cpp
@@ -116,11 +116,15 @@ void init_MultiFab(py::module &m)
         .def("is_nodal",
              py::overload_cast< int >(&FabArrayBase::is_nodal, py::const_))
 
-        .def_property_readonly("nComp", &FabArrayBase::nComp)
-        .def_property_readonly("num_comp", &FabArrayBase::nComp)
-        .def_property_readonly("size", &FabArrayBase::size)
+        .def_property_readonly("nComp", &FabArrayBase::nComp,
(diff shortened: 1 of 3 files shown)


src/Base/MultiFab.cpp
```diff
@@ -193,22 +337,69 @@ void init_MultiFab(py::module &m)
...
+    py_MultiFab
         .def(py::init< const BoxArray&, const DistributionMapping&, int, int,
-                       MFInfo const &, FabFactory<FArrayBox> const & >())
+                       MFInfo const &, FabFactory<FArrayBox> const & >(),
+            py::arg("bxs"), py::arg("dm"), py::arg("ncomp"), py::arg("ngrow"),
+#ifdef AMREX_STRICT_MODE
+            py::arg("info"), py::arg("factory"),
+#else
+            py::arg("info") = MFInfo(),
+            py::arg("factory") = FArrayBoxFactory(),
```
ax3l:
```
ImportError: arg(): could not convert default argument 'factory: amrex::DefaultFabFactory<amrex::FArrayBox>' in method '<class 'amrex.space1d.amrex_1d_pybind.MultiFab'>.__init__' into a Python object (type not registered yet?)
```

ax3l: I think in Python, such default constructor _calls_ do generally not work (and would only create an object once at import time). Will remove those here.

PR #293, opened 2024-04-11
Add overload for `sum_unique`

RemiLehe:
This is a follow-up on this PR: https://github.com/AMReX-Codes/amrex/pull/3871

### DIFF
--- a/src/Base/MultiFab.cpp
+++ b/src/Base/MultiFab.cpp
@@ -507,18 +507,30 @@ void init_MultiFab(py::module &m)
         .def("norm2", py::overload_cast< Vector<int> const & >(&MultiFab::norm2, py::const_))
 
         /* simple math */
+
         .def("sum",
              //py::overload_cast< int, bool >(&MultiFab::sum, py::const_),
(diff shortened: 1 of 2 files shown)


src/Base/MultiFab.cpp
```diff
@@ -507,19 +507,31 @@ void init_MultiFab(py::module &m)
         .def("norm2", py::overload_cast< Vector<int> const & >(&MultiFab::norm2, py::const_))
 
         /* simple math */
+
         .def("sum",
              //py::overload_cast< int, bool >(&MultiFab::sum, py::const_),
              [](MultiFab const & mf, int comp , bool local) { return mf.sum(comp, local); },
```
RemiLehe: I also tried the following change (as a preparation step, before adding the overload for `sum` that takes a `region` argument), but it surprisingly fails to compile
```suggestion
        .def("sum",
             py::overload_cast< int, bool >(&MultiFab::sum, py::const_),
```

ax3l: That is probably some issue because the base class of MultiFab also defines `sum` overloads. We can simply use lambdas.

ax3l (approved):
Thank you! :)

PR #300, opened 2024-04-25
`Geometry`: Doc Strings

ax3l:
Add more argument names and doc strings.

Breaking: Use properties more and avoid equivalent overloads.


RTSandberg (approved):
The doc strings and conversion from member function to attributes looks good to me

PR #301, opened 2024-04-25
`MultiFab`: Static to Member Math Methods

### LINKED ISSUE #296: MultiFab: Only Member Functions
Make the static freestanding functions in MultiFab only member functions, defaulting the `dst` argument to `self`.

cc @RemiLehe

ax3l:
Replace `static` member functions that with regular member functions that take store the result/destination in `self`. (Assume `dst` is `self`.)

- [x] `FabArray<FArrayBox>`
- [x] `MultiFab`

Close #296

### DIFF
--- a/src/Base/MultiFab.cpp
+++ b/src/Base/MultiFab.cpp
@@ -12,6 +12,7 @@
 #include <AMReX_FabArrayBase.H>
 #include <AMReX_FabFactory.H>
 #include <AMReX_MultiFab.H>
+#include <AMReX_iMultiFab.H>
 
 #include <memory>
 #include <string>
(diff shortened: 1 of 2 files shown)


src/Base/MultiFab.cpp
```diff
@@ -745,108 +791,136 @@ void init_MultiFab(py::module &m)
         )
         //.def_static("dot", py::overload_cast< iMultiFab const&, const MultiFab&, int, MultiFab const&, int, int, int, bool >(&MultiFab::Dot))
```
ax3l: I think for the most general case for `dot`, this needs to return a new MF, no?

ax3l: Ah wait, this returns a `Real`. We can still make this a member function.
I was confused and thought of another operation, lost in translation

ax3l:
I will go ahead and merge this, because I have a massive restructuring PR to add support for `iMultiFab` that depends on these changes: #410

PR #307, opened 2024-05-01
Fix CMake: CUDA Compile Features

ax3l:
Fix the requested compiler features for CUDA compiles. Should not ask for CXX language features in that case.


ax3l:
Uhoh, that could be a more general breakage in latest CMake releases and we have to update our overall AMReX CUDA CMake treatment finally...

ax3l:
Update: technically, the old way did work as well, we just do usually not ask CXX language features for CUDA targets (e.g., in WarpX).

PR #333, opened 2024-06-19
Fix ccache for macos actions

WeiqunZhang:
* Need to ignore time macros. Brew seems to inject them from some dependent packages.

* Bump up cache action's version


ax3l (approved):
Thank you! :tada:

PR #339, opened 2024-07-16
Fix CMake IPO Control

ax3l:
`pybind11::lto` is only defined if `CMAKE_INTERPROCEDURAL_OPTIMIZATION` is not set in [pybind11Common.cmake](https://github.com/pybind/pybind11/blob/v2.12.0/tools/pybind11Common.cmake#L397-L403). [Package managers like Spack use the latter.](https://github.com/spack/spack/pull/18374)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,9 +68,13 @@ pyamrex_set_default_install_dirs()
 
 # Options and Variants ########################################################
 #
+set(_pyAMReX_IPO_DEFAULT ON)
+if(DEFINED CMAKE_INTERPROCEDURAL_OPTIMIZATION)
+    set(_pyAMReX_IPO_DEFAULT ${CMAKE_INTERPROCEDURAL_OPTIMIZATION})
+endif()
 option(pyAMReX_IPO
     "Compile with interprocedural optimization (IPO) / link-time optimization (LTO)"
-    ON
+    ${_pyAMReX_IPO_DEFAULT}
 )
 option(pyAMReX_INSTALL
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -135,7 +139,13 @@ foreach(D IN LISTS AMReX_SPACEDIM)
...
     target_link_libraries(pyAMReX_${D}d PRIVATE pybind11::module pybind11::windows_extras)
     if(pyAMReX_IPO)
-        target_link_libraries(pyAMReX_${D}d PRIVATE pybind11::lto)
+        if(NOT DEFINED CMAKE_INTERPROCEDURAL_OPTIMIZATION)
+            # conditionally defined target in pybind11
+            # https://github.com/pybind/pybind11/blob/v2.12.0/tools/pybind11Common.cmake#L397-L403
+            target_link_libraries(pyAMReX_${D}d PRIVATE pybind11::lto)
+        else()
+            pyamrex_enable_IPO(pyAMReX_${D}d)
+        endif()
```
ax3l: Simpler branches:
```suggestion
        if(DEFINED CMAKE_INTERPROCEDURAL_OPTIMIZATION)
            pyamrex_enable_IPO(pyAMReX_${D}d)
        else()
            # conditionally defined target in pybind11
            # https://github.com/pybind/pybind11/blob/v2.12.0/tools/pybind11Common.cmake#L397-L403
            target_link_libraries(pyAMReX_${D}d PRIVATE pybind11::lto)
        endif()
```

PR #342, opened 2024-07-22
Add Particle I/O: restart & restart_checkpoint

JBlaschke:
Adding particle I/O. Working on proper tests now.

This will also read FHDeX particles. 

@ax3l let me know if you have early feedback


src/Particle/ParticleContainer.H
```diff
@@ -312,8 +312,34 @@ void make_ParticleContainer_and_Iterators (py::module &m, std::string allocstr)
...
+
         // void Restart (const std::string& dir, const std::string& file);
+        .def("restart",
+            [](
+                ParticleContainerType & pc,
+                std::string const & dir,
+                std::string const & file
+            ){
+                return pc.Restart(dir, file);
+            },
```
ax3l: Does this shortcut work, too?
```suggestion
        .def("restart",
            py::overload_cast<std::string const &, std::string const &>(&ParticleContainerType::Restart),
```

src/Particle/ParticleContainer.H
```diff
@@ -312,8 +312,34 @@ void make_ParticleContainer_and_Iterators (py::module &m, std::string allocstr)
...
         // void Restart (const std::string& dir, const std::string& file, bool is_checkpoint);
+        .def("restart_checkpoint",
+            [](
+                ParticleContainerType & pc,
+                std::string const & dir,
+                std::string const & file,
+                bool is_checkpoint
+            ){
+                return pc.Restart(dir, file, is_checkpoint);
+            },
```
ax3l: Potential shortcut to try:
```suggestion
            py::overload_cast<std::string const &, std::string const &, bool>(&ParticleContainerType::Restart),
```

tests/test_plotfileparticledata.py
```diff
@@ -0,0 +1,50 @@
+from dataclasses import dataclass
+
+import amrex.space3d as amr
+
+if not amr.initialized():
+    amr.initialize([])
```
ax3l: The `tests/conftest.py` file does that.

tests/test_plotfileparticledata.py
```diff
@@ -0,0 +1,50 @@
...
+    z: float
+
+    ib_id: int
+    idx: int
+
+
+def load_data(plot_file_name):
+    plt = amr.PlotFileData(plt_file_name)
+
+    probDomain = plt.probDomain(0)
```
ax3l: For pytest, you an start those with `def test_...` to be picked up.

if you like to run tests as standalone, you could add a main to the file, but usually we run them via pytest.

ax3l (commented):
Looks great so far! :+1:

BenWibking:
Does a separate particle container need to be created for each app? We have this (+more to come):
https://github.com/quokka-astro/quokka/blob/c37cfb21266fb793d2ff39d865a3f2174a3912e9/src/particles/CICParticles.hpp#L25

BenWibking:
The real problem is that we will have some ParticleContainers with a variable number of real components for each simulation...

ax3l:
> Does a separate particle container need to be created for each app? We have this (+more to come):
https://github.com/quokka-astro/quokka/blob/c37cfb21266fb793d2ff39d865a3f2174a3912e9/src/particles/CICParticles.hpp#L25
> The real problem is that we will have some ParticleContainers with a variable number of real components for each simulation...

Yes, so far.
I see two ways to solve this:
a) expose the header for PC creation, so you can specialize for your code downstream or
b) more forward looking, codes could try to use a minimal set of compile-time components (e.g., ND positions only in pure SoA format) and instead use runtime components for all other attributes. That way, we only have to specialize one base class that everyone ca use. I doubt there is a significant performance impact, but we need to evaluate that in applications.

Another change I will put into BLAST codes is to use the Polymorphic particle container, where Arena is not a template parameter. On the PC level, having the Arena a C++ template appears unnecessary to me and should not have any performance benefits (in symmetry, the MultiFabs also have it only as a runtime parameter).

JBlaschke:
I'm currently stalled on a head-scratcher: when I run `test_aos.py` on its own, it works fine. If I run it in `pytest` I get:
```
tests/test_aos.py::test_aos_init amrex::Error::0::ParmParse::Initialize(): already initialized! !!!
```
Most strange ... any advice?

ax3l:
The issue is a double init. You can remove this:
https://github.com/AMReX-Codes/pyamrex/pull/342#discussion_r1696132626

We init and finalize amrex in a [fixture](https://docs.pytest.org/en/6.2.x/fixture.html) for all tests here:
https://github.com/AMReX-Codes/pyamrex/blob/development/tests/conftest.py
so this is not needed again in the test file.

To run only your latest test while you develop, [run this](https://github.com/AMReX-Codes/pyamrex?tab=readme-ov-file#test):
```bash
python3 -m pytest tests/test_plotfileparticledata.py
```
the tests are not intended to be run without the pytest module, but can be sub-selected for productivity to feel like a standalone script.

tests/test_plotfileparticledata.py
ax3l: Remove the lines to get rid of the `already initialized` error:
```suggestion
```

JBlaschke: Oh! Pytest's verbose mode was being highly misleading: it must have crashed just after printing `tests/test_aos.py` which made it look like the error was coming from there, when it was coming from somewhere else.

So why does this cause an error? If `amr.initialize` is being guarded. I like this approach because then a relatively inexperienced user can run the test without needing to know what pytest does behind the scenes.

I.e. I would prefer to have similar behaviour when it comes to initializing as `mpi4py`

JBlaschke: Ah! So the problem was that all tests are imported as modules. So my test was initializing AMReX before pytest was. 

* I added guard against this here: https://github.com/AMReX-Codes/pyamrex/pull/342/commits/fa20a6fb89038fe1131a6f0a3c60773648cf1c01

* And I am guarding against initializing pyamrex before the fixture does here: https://github.com/AMReX-Codes/pyamrex/pull/342/commits/04c01427713f7f8b4e5c1b891ad9749c7886bb59

This lets a user build off the test as an example without needing to know that pytest autmatically initialized amrex.

What do you think @ax3l ?

ax3l: Thanks! Yes, when I learned pytest the other year I was also having the same urge initially. Nonetheless by now, since this adds more overhead for tests to maintain and they can literally be called with the documented one-liner like an individual script + the unification of options we want for all tests that we put in `conftest.py` I would keep it simple, pytest only.

ax3l:
It looks like when building in debug mode (with more checks), the test still fails.

JBlaschke:
Any idea why the tests fail? Seems to depend on compiler version... Which is never a good sign...

tests/test_plotfileparticledata.py
```diff
@@ -0,0 +1,163 @@
...
+
+    particles_tile_ct = 0
+    # assign some values to runtime components
+    for lvl in range(pc.finest_level + 1):
+        for pti in pc.iterator(pc, level=lvl):
+            aos = pti.aos()
+            aos_numpy = aos.to_numpy(copy=False)
+            for i, p in enumerate(aos_numpy):
+                Rpart[i + particles_tile_ct].to_soa(p)
+            particles_tile_ct += len(aos_numpy)
```
ax3l, line 73: I think there is an issue in this block, returning before it keeps the pc `.OK()`.

Specifically the `Rpart[i + particles_tile_ct].to_soa(p)` seems to cause the issue.

ax3l: This fixes the `pc.OK()` check for me:
```suggestion
    particles_tile_ct = 0
    # assign some values to runtime components
    for lvl in range(pc.finest_level + 1):
        for pti in pc.iterator(pc, level=lvl):
            aos = pti.aos()
            aos_numpy = aos.to_numpy(copy=True)
            for i, p in enumerate(aos_numpy):
                Rpart[i + particles_tile_ct].to_soa(p)
            particles_tile_ct += len(aos_numpy)
```

ax3l: Ah, I think the issue might be that some particles are out of bounds after the assignment here for their current boxes?

I added a `pc.redistribute()` before returning to make it pass, but we need to check the particles are not just removed now:
```suggestion
    particles_tile_ct = 0
    # assign some values to runtime components
    for lvl in range(pc.finest_level + 1):
        for pti in pc.iterator(pc, level=lvl):
            aos = pti.aos()
            aos_numpy = aos.to_numpy(copy=False)
            for i, p in enumerate(aos_numpy):
                Rpart[i + particles_tile_ct].to_soa(p)
            particles_tile_ct += len(aos_numpy)

    pc.redistribute()
    return pc     
```

ax3l:
Please see comments from yesterday. It fails in debug mode, where AMReX does more checks for validity. Likely because the added positions do not line up with the domain decomposition.

tests/test_plotfileparticledata.py
```diff
@@ -0,0 +1,166 @@
...
+
+    pc.redistribute()
+    return pc     
+
+    return pc
```
ax3l:
```suggestion
```

src/Particle/ParticleContainer.cpp
```diff
@@ -59,6 +60,7 @@ void init_ParticleContainer(py::module& m) {
     init_ParticleContainer_HiPACE(m);
     init_ParticleContainer_ImpactX(m);
     init_ParticleContainer_WarpX(m);
+    init_ParticleContainer_FHDeX(m);
```
ax3l:
```suggestion
    init_ParticleContainer_FHDeX(m);
    init_ParticleContainer_HiPACE(m);
    init_ParticleContainer_ImpactX(m);
    init_ParticleContainer_WarpX(m);
```

src/Particle/ParticleContainer.cpp
```diff
@@ -36,6 +36,7 @@ namespace
 void init_ParticleContainer_HiPACE(py::module& m);
 void init_ParticleContainer_ImpactX(py::module& m);
 void init_ParticleContainer_WarpX(py::module& m);
+void init_ParticleContainer_FHDeX(py::module& m);
```
ax3l:
```suggestion
void init_ParticleContainer_FHDeX(py::module& m);
void init_ParticleContainer_HiPACE(py::module& m);
void init_ParticleContainer_ImpactX(py::module& m);
void init_ParticleContainer_WarpX(py::module& m);
```

src/Particle/CMakeLists.txt
```diff
@@ -5,5 +5,6 @@ foreach(D IN LISTS AMReX_SPACEDIM)
         ParticleContainer_HiPACE.cpp
         ParticleContainer_ImpactX.cpp
         ParticleContainer_WarpX.cpp
+        ParticleContainer_FHDeX.cpp
```
ax3l:
```suggestion
        ParticleContainer_FHDeX.cpp
        ParticleContainer_HiPACE.cpp
        ParticleContainer_ImpactX.cpp
        ParticleContainer_WarpX.cpp
```

tests/test_plotfileparticledata.py
```diff
@@ -0,0 +1,160 @@
...
+    # assign some values to runtime components
+    for lvl in range(pc.finest_level + 1):
+        for pti in pc.iterator(pc, level=lvl):
+            aos = pti.aos()
+            aos_numpy = aos.to_numpy(copy=False)
+            for i, p in enumerate(aos_numpy):
+                Rpart[i + particles_tile_ct].to_soa(p)
+            particles_tile_ct += len(aos_numpy)
+
+    pc.redistribute()
```
ax3l, line 75: @JBlaschke This fixed the `pc.OK()` issue that you saw when running AMReX in Debug mode. Please double check (maybe with `print` debugging) if this does what you expect to do or if you geometry needs adjustment, if the `pc.redistribute` removes particles.

ax3l (approved):
Thank you, Johannes! :rocket: :sparkles:

src/Particle/ParticleContainer_FHDeX.cpp
```diff
@@ -0,0 +1,16 @@
...
+#include "ParticleContainer.H"
+
+#include <AMReX_Particle.H>
+#include <AMReX_ParticleTile.H>
+
+
+void init_ParticleContainer_FHDeX(py::module& m) {
+    using namespace amrex;
+
+    make_ParticleContainer_and_Iterators<Particle<16, 4>, 0, 0>(m);
```
ax3l, line 15: @JBlaschke do you need this for 1D, 2D and 3D AMReX builds or only for 3D?

PR #345, opened 2024-07-23
Add overload_cast to IntVect::dim3()

AlexanderSinn:
In https://github.com/AMReX-Codes/amrex/pull/4016 I want to add another overload to `dim3()` which will require an overload_cast here.


ax3l (approved):
Thank you! :)

PR #351, opened 2024-08-10
CI: Ruff replaced Tools

ax3l:
https://docs.astral.sh/ruff/ replaces a lot of linter and formatter tools, remove duplication for stability and less noise.


ax3l:
Small regression: unstable sort of imports pushes multiple rotating commits to `development`
https://github.com/astral-sh/ruff/issues/12872

PR #360, opened 2024-08-28
CI: Save PR Number Once

ax3l:
Only run the `save_pr_number` script once per PR.

@WeiqunZhang, should this run for all our workflows per PR
- [ ] last
- [ ] first
- [x] at any time in parallel to running tests
- [ ] only on success of all tests
- [ ] ...?

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -120,18 +120,3 @@ jobs:
         uses: github/codeql-action/upload-sarif@v3
         with:
           sarif_file: sarif-results/${{ matrix.language }}.sarif
-
-  save_pr_number:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -120,18 +120,3 @@ jobs:
         uses: github/codeql-action/upload-sarif@v3
         with:
           sarif_file: sarif-results/${{ matrix.language }}.sarif
-
-  save_pr_number:
```
ax3l, line 124: This is also in `ci.yml`, which calls this workflow script after other prerequisite runs have succeeded.

WeiqunZhang (approved):

PR #366, opened 2024-09-06
EB: On by Default in Superbuild

ax3l:
Now that AMReX EBs are fully runtime controlled, we will default use them in dependent projects like WarpX and can default to them here as well.

X-ref: https://github.com/ECP-WarpX/WarpX/pull/4865

### DIFF
--- a/cmake/dependencies/AMReX.cmake
+++ b/cmake/dependencies/AMReX.cmake
@@ -66,7 +66,8 @@ macro(find_amrex)
     elseif(NOT pyAMReX_amrex_internal)
         message(STATUS "Searching for pre-installed AMReX ...")
         # https://amrex-codes.github.io/amrex/docs_html/BuildingAMReX.html#importing-amrex-into-your-cmake-project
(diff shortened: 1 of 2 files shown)


cmake/dependencies/AMReX.cmake
```diff
@@ -65,7 +66,7 @@ macro(find_amrex)
     elseif(NOT pyAMReX_amrex_internal)
         message(STATUS "Searching for pre-installed AMReX ...")
         # https://amrex-codes.github.io/amrex/docs_html/BuildingAMReX.html#importing-amrex-into-your-cmake-project
-        find_package(AMReX 24.09 CONFIG REQUIRED COMPONENTS PARTICLES PIC)
+        find_package(AMReX 24.09 CONFIG REQUIRED COMPONENTS EB PARTICLES PIC)
```
ax3l:
```suggestion
        # not strictly required yet to compile pyAMReX: EB
        find_package(AMReX 24.09 CONFIG REQUIRED COMPONENTS PARTICLES PIC)
```

PR #370, opened 2024-09-27
Implement multifab wrapper functionality from WarpX

dpgrote:
This adds the code from the WarpX fields.py that allows global indexing of MultiFabs. This is done by adding `__getitem__` and `__setitem__` methods for the MultiFab object. 

Note that this incorporates the changes from WarpX PR [#4370](https://github.com/ECP-WarpX/WarpX/pull/4370) that had not been merged in.

This currently only does the get/set without including the ghost cells. What is needed is a way of specifying  inclusion of the ghost cells. Perhaps this?
```py
mf = MultiFab()
mf[:,5,6]  # without ghost cells
mf[:,5,6] = 2.  # without ghost cells
mf.with_ghosts([slice(None),5,6])  # with ghost cells (in the first dimension)
mf.with_ghosts([slice(None),5,6], 2.)  # with ghost cells
```
Maybe this, using imaginary numbers to refer to ghost cells?
```py
mf[-1j,5,6]  # the first ghost cell on the lower boundary
mf[+1j,5,6]  # the first ghost cell on the upper boundary
mf[(),5,6]  # the full first dimension including ghosts
mf[:,5,6]  # the first dimension without ghosts
```
I prefer this method.

### DIFF
--- a/src/amrex/extensions/MultiFab.py
+++ b/src/amrex/extensions/MultiFab.py
@@ -1,12 +1,21 @@
 """
 This file is part of pyAMReX
 
-Copyright 2023 AMReX community
-Authors: Axel Huebl
+Copyright 2024 AMReX community
+Authors: Axel Huebl, David Grote
 License: BSD-3-Clause-LBNL
 """
 
 from .Iterator import next
+
+import numpy as np
+
+try:
+    from mpi4py import MPI as mpi
+    comm_world = mpi.COMM_WORLD
+    npes = comm_world.Get_size()
+except ImportError:
+    npes = 1
 
 
 def mf_to_numpy(self, copy=False, order="F"):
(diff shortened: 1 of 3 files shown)


ax3l:
Hi @dpgrote,

That is fantastic, thanks a lot for porting this!

I think in symmetry with AMReX functions like `Box::grow()`, the syntax `.with_ghosts` is clearer. Could also be `.grow` if we want, but either way is great.

We could even add default arguments to `=nGrowVect()` for the function so it can be called `.with_ghosts()` and get the ghosts.

ax3l:
@dpgrote This is awesome. To add CI and show usage examples, can you please add a few new examples in `tests/`?

This then makes it very easy to add add a new `tab` in fields, using the snippet from the test then in the user-facing doc page:
- user-facing: https://pyamrex.readthedocs.io/en/latest/usage/compute.html#fields
- source annotation example: https://github.com/AMReX-Codes/pyamrex/blob/1c66690f83244196c5655293f1381303a7d1589d/tests/test_multifab.py#L47-L64
- https://github.com/AMReX-Codes/pyamrex/blob/development/docs/source/usage/compute.rst

src/amrex/extensions/MultiFab.py
```diff
@@ -173,6 +183,496 @@ def copy_multifab(amr, self):
...
+            global_shape[0:0] = [1]
+        if not isinstance(ii[1], slice):
+            global_shape[1:1] = [1]
+        if not isinstance(ii[2], slice):
+            global_shape[2:2] = [1]
+        if not isinstance(ic, slice) or len(global_shape) < 4:
+            global_shape[3:3] = [1]
+        value3d.shape = global_shape
+
+        if libwarpx.libwarpx_so.Config.have_gpu:
```
ax3l:
```suggestion
        if amr.Config.have_gpu:
```

src/amrex/extensions/MultiFab.py
```diff
@@ -173,6 +183,496 @@ def copy_multifab(amr, self):
...
+            # check if cupy is available for use
+            xp, cupy_status = load_cupy()
+            if cupy_status is not None:
+                libwarpx.amr.Print(cupy_status)
```
ax3l:
```suggestion
                amr.Print(cupy_status)
```

src/amrex/extensions/MultiFab.py
```diff
@@ -173,6 +183,496 @@ def copy_multifab(amr, self):
...
+    indices will refer to the lower guard cells.
+
+    Parameters
+    ----------
+    index : integer, or sequence of integers or slices, or Ellipsis
+        The slice to set
+    value : scalar or array
+        Input value to assign to the specified slice of the MultiFab
+    """
+    # Temporary value until fixed
```
ax3l:
```suggestion
    import inspect

    amr = inspect.getmodule(self)

    # Temporary value until fixed
```

src/amrex/extensions/MultiFab.py
```diff
@@ -173,6 +183,496 @@ def copy_multifab(amr, self):
...
+    # are removed, return a scalar (that's what the [()] does)
+    return result_global.squeeze()[()]
+
+
+def __setitem__(self, index, value):
+    """Sets slices of a decomposed array using global indexing.
+    The shape of the input object depends on the number of arguments specified, which can
+    be from none to all three.
+    This allows negative indexing, though with ghosts cells included, the first n-ghost negative
+    indices will refer to the lower guard cells.
```
ax3l: For `__getitem__` and `__setitem__`, I would add a bit more info of the form:
```suggestion
    indices will refer to the lower guard cells.
    
    Note: in MPI-parallel contexts, this triggers a...
```
- point-to-point? global? MPI communication
- are the setter/getter independent or collective operations in MPI? (e.g., does every rank need to set the same value? will it hang if a single rank sets a value in an `if`?)

src/amrex/extensions/MultiFab.py
```diff
@@ -173,6 +183,496 @@ def copy_multifab(amr, self):
...
+
+
+def __getitem__(self, index):
+    """Returns slice of the MultiFab using global indexing.
+    The shape of the object returned depends on the number of ix, iy and iz specified, which
+    can be from none to all three. Note that the values of ix, iy and iz are
+    relative to the fortran indexing, meaning that 0 is the lower boundary
+    of the whole domain, and in fortran ordering, i.e. [ix,iy,iz].
+    This allows negative indexing, though with ghosts cells included, the first n-ghost negative
+    indices will refer to the lower guard cells.
```
ax3l: For `__getitem__` and `__setitem__`, I would add a bit more info of the form:
```suggestion
    indices will refer to the lower guard cells.

    Note: in MPI-parallel contexts, this triggers a...
```
Please see more details below on `__setitem__`:

dpgrote:
> I think in symmetry with AMReX functions like `Box::grow()`, the syntax `.with_ghosts` is clearer. Could also be `.grow` if we want, but either way is great.
> 
> We could even add default arguments to `=nGrowVect()` for the function so it can be called `.with_ghosts()` and get the ghosts.

How about this, add the method `mf.alias_with_ghosts()` that returns an alias of the MF with the boxes grown by the number of ghost cells. Then the `__getitem__` and `__setitem__` can be use directly.

Note that there is a subtlety in `__setitem__` regarding ghosts. With `include_ghosts` False, only valid cells are set. When True, the ghost cells of the individual blocks are set as well (as if `FillBoundary` had been called). Perhaps the internal ghost cells should always be set?

src/amrex/extensions/MultiFab.py
```diff
@@ -1,13 +1,23 @@
...
+Copyright 2024 AMReX community
+Authors: Axel Huebl, David Grote
 License: BSD-3-Clause-LBNL
 """
 
+import numpy as np
+
 from .Iterator import next
 
+try:
```
ax3l: We need to put this into a function that has access to `amr`, so we only attempt an import if:
```py
if amr.Config.have_mpi:
```

Otherwise, this can cause issues in scenarios, e.g., running serially on Cray machines.

src/amrex/extensions/MultiFab.py
```diff
@@ -1,13 +1,23 @@
...
+    from mpi4py import MPI as mpi
+
+    comm_world = mpi.COMM_WORLD
```
ax3l: While we have access to `amr`, we can replace `mpi.COMM_WORLD` with `amr.ParallelDescriptor.Communicator()` later on.

I did not expose that element yet, but we already created an mpi4py compatible `MPI_Communicator` for the MPMD logic in pyAMReX. This can be a follow-up PR.

dpgrote:
The more I think about this (and work on implementing something), the more I like the idea of using imaginary numbers for the guard cells. The idea of using negative numbers for the lower ghost cells breaks the standard python usage, since for every sequence type, negative indices are always relative to the end of the sequence. It is asymmetric since there then no nice way of referring to the upper ghost cells. The imaginary indices are a very concise and unambiguous way is specifying the ghosts, both lower and upper. I'll implement this and provide examples to show what it's like.

ax3l:
So to get a plane with ghost cells, something like this? :)
```py
mf[-2j:+2j, 0, 0]  # include 2 ghost cells in the first plane of the global domain
```
or is that
```py
mf[-2j:-1+3j, 0, 0]  # include 2 ghost cells in the first plane of the global domain
```

dpgrote:
> So to get a plane with ghost cells, something like this? :)
> 
> ```python
> mf[-2j:+2j, 0, 0]  # include 2 ghost cells in the first plane of the global domain
> ```
> 
> or is that
> 
> ```python
> mf[-2j:-1+3j, 0, 0]  # include 2 ghost cells in the first plane of the global domain
> ```

It would be
```
mf[-2j:0, 0, 0]
```
The `-2j` refers to the 2nd ghost cell. `0` refers to the first valid cell, remembering python slice range that goes up to but does not include the end of the slice range.

If you want all of the cells, the valid cells plus the two guard cells on each side, then it would be
```
mf[-2j:3j,0,0]
```
One thing that is missing is a clean way to specify and open ended range on one side including the ghost cells.

ax3l (approved):
This is awesome and impressive, thank you! :rocket: :sparkles:

PR #378, opened 2024-10-09
[WIP] CMake: Tarball Support

ax3l:
This speeds up the configure step from 20.1sec to 6.5sec by just downloading a shallow copy of AMReX & pybind11.

- [ ] check AMReX version output during configure and ensure build version logic in AMReX is still sensible (e.g., output during AMReX initialize)
- [ ] extend the update script to simply update tar sha and git branch at the same time for AMReX & pybind11 updates  #474

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,6 +80,9 @@ option(pyAMReX_INSTALL
     "Enable install targets for pyAMReX"
     ON
 )
+option(pyAMReX_SUPERBUILD
+    "Download & build extra dependencies"
+    ON)
 
 # change the default build type to Release (or RelWithDebInfo) instead of Debug
 set_default_build_type("Release")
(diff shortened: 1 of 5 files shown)


ax3l:
@EZoni @WeiqunZhang this PR is _not_ urgent and I can take it on when I am back from parental leave. I want to introduce it consistently in pyAMReX/WarpX/ImpactX/HiPACE during the same release cycle. Can wait for 2025.

I already use the same trick in openPMD-api 0.16.0+
As an alternative approach, one could do shallow (`--depth 1`) git clones, but unfortunately, the CMake git fetcher does not support shallow copies.

ax3l:
Hi @EZoni,

Do you mind adding the WarpX `Tools/Release/update_dependencies.py` script (simply in `/`) to pyAMReX, too?
I would then improve that script in the PR here to automatically calculate checksums for the tarballs.

PR #380, opened 2024-10-10
Print IntVect values in repr

dpgrote:
This does what the title says. Note that the RealVect already does this.

### DIFF
--- a/src/Base/IntVect.cpp
+++ b/src/Base/IntVect.cpp
@@ -28,8 +28,10 @@ namespace
         py::class_< iv_type > py_iv(m, iv_name.c_str());
         py_iv
             .def("__repr__",
-                 [iv_name](const iv_type&) {
-                     return "<amrex." + iv_name + ">";
+                 [iv_name](const iv_type& iv) {
+                     std::stringstream s;
+                     s << iv;
+                     return "<amrex." + iv_name + s.str() + ">";
                  }
             )
             .def("__str",


src/Base/IntVect.cpp
```diff
@@ -28,8 +28,10 @@ namespace
         py::class_< iv_type > py_iv(m, iv_name.c_str());
         py_iv
             .def("__repr__",
-                 [iv_name](const iv_type&) {
-                     return "<amrex." + iv_name + ">";
+                 [iv_name](const iv_type& iv) {
+                     std::stringstream s;
+                     s << iv;
+                     return "<amrex." + iv_name + s.str() + ">";
```
ax3l, line 34: Would this be more natural?
I forgot what the `<<` for `iv` implements
```suggestion
                     return "<amrex." + iv_name + ", " + s.str() + ">";
```

ax3l:
@atmyers can you please take the review over? :pray:

src/Base/IntVect.cpp
dpgrote: Either way is fine. As I have it, it would print something like
`<amrex.IntVect(10,20,30)>`
versus
`<amrex.IntVect, (10,20,30)>`

ax3l: Thanks, got it.

ax3l (approved):

PR #388, opened 2024-11-01
Add wrapper for VisMF

dpgrote:
I need to read in a MultiFab so added this.

### DIFF
--- a/src/Base/CMakeLists.txt
+++ b/src/Base/CMakeLists.txt
@@ -33,6 +33,7 @@ foreach(D IN LISTS AMReX_SPACEDIM)
         Utility.cpp
         Vector.cpp
         Version.cpp
+        VisMF.cpp
     )
 
     if (AMReX_MPI)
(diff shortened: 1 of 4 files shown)


WeiqunZhang:
There is a potential issue. The VisMF::Read wrapper always returns a freshly made MultiFab. The problem of that is if a users calls the wrapper twice to read two MulitiFabs, there is no guarantee that the two will have the same DistributionMapping. That may not be what the user expects. So it might be better to also add the option of reading the data into a preexisting MultiFab.

dpgrote:
If the two MultiFabs files are from MultiFabs with the same box array layout and distribution map, shouldn't they have the same distribution map when they are read back in (or at least the two should have the same as each other)?

I added a second version of Read that takes a MultiFab as an argument.

WeiqunZhang:
The DistributionMapping is not saved in the VisMF files. For flexibility in load balancing, using the same boxarray to build distributionmapping does not always produce the same result.

However, we can probably make an exception in VisMF::Read to make it less error prone for the users.

WeiqunZhang (approved):

PR #417, opened 2025-03-04
ParmParse: `pretty_print_table`

ax3l:
Print to python output.

### DIFF
--- a/src/Base/ParmParse.cpp
+++ b/src/Base/ParmParse.cpp
@@ -9,6 +9,7 @@
 #include <AMReX_IntVect.H>
 #include <AMReX_ParmParse.H>
 
+#include <iostream>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


ax3l:
@WeiqunZhang @atmyers it would also be nice to implement a `.to_dict()` method for `ParmParse`, but I do not know how to automatically find out if something is a float/int/str attribute. Any idea how we could systematically guess scalar/array float/int/str type of each `PP_Entry` of `ParmParse`?
Do we store type hints once a user/program adds a value into `ParmParse`?

WeiqunZhang:
The values in PP_Entry are strings. A value of say `3` can be interpreted as either bool, int, long, float, or double, as long as it's convertible.

ax3l:
Yep, I wouldn't know if there is a sensible order (hierarchy) to cast them. Could we simply store a type hint as well? Whenever a value gets added in the first place we know the type, don't we?

Update: ah no, not if we get it from CLI or a text file of course :sweat_smile: But for anything programmatically added (i.e., C++ & Python added) we could actually store a hint with.

PR #430, opened 2025-03-20
Tests: `amrex.the_arena_init_size=0`

ax3l:
`amrex.the_arena_init_size=0` is default on CPU but not GPU. In order to run tests in parallel on the same GPU, we need to change the default so the first test does not allocate 3/4th of the memory of the GPU (to speed up later allocations in a pre-allocated heap).


EZoni:
Merging `development` to rerun CI after the fixes in #436 and #441.

Companion PR in WarpX for reference: https://github.com/BLAST-WarpX/warpx/pull/5785.

EZoni (approved):

PR #435, opened 2025-04-16
Add wrappers to invalidate particles and set id/cpu from Python

atmyers:
To see how this is used, see [tests/test_particleTile.py](https://github.com/AMReX-Codes/pyamrex/pull/435/files#diff-ecc0338faa7592dbb084e945148b2aaa8f982e06dbc8cdcf72dcdf751a02dacb)

### DIFF
--- a/cmake/dependencies/AMReX.cmake
+++ b/cmake/dependencies/AMReX.cmake
@@ -86,7 +86,7 @@ option(pyAMReX_amrex_internal "Download & build AMReX" ON)
 set(pyAMReX_amrex_repo "https://github.com/AMReX-Codes/amrex.git"
     CACHE STRING
     "Repository URI to pull and build AMReX from if(pyAMReX_amrex_internal)")
-set(pyAMReX_amrex_branch "25.04"
+set(pyAMReX_amrex_branch "development"
     CACHE STRING
     "Repository branch for pyAMReX_amrex_repo if(pyAMReX_amrex_internal)")
 
--- a/cmake/dependencies/AMReX.cmake
+++ b/cmake/dependencies/AMReX.cmake
@@ -86,7 +86,7 @@ option(pyAMReX_amrex_internal "Download & build AMReX" ON)
 set(pyAMReX_amrex_repo "https://github.com/AMReX-Codes/amrex.git"
     CACHE STRING
     "Repository URI to pull and build AMReX from if(pyAMReX_amrex_internal)")
-set(pyAMReX_amrex_branch "development"
+set(pyAMReX_amrex_branch "793ea9f717590d66d178a86298b82aac244b77a7"
     CACHE STRING
     "Repository branch for pyAMReX_amrex_repo if(pyAMReX_amrex_internal)")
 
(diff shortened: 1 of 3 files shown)


cmake/dependencies/AMReX.cmake
```diff
@@ -86,7 +86,7 @@ option(pyAMReX_amrex_internal "Download & build AMReX" ON)
 set(pyAMReX_amrex_repo "https://github.com/AMReX-Codes/amrex.git"
     CACHE STRING
     "Repository URI to pull and build AMReX from if(pyAMReX_amrex_internal)")
-set(pyAMReX_amrex_branch "25.04"
+set(pyAMReX_amrex_branch "development"
```
EZoni: I thought we usually set a commit hash here, rather than the name of the branch. But this works too, of course. It will be soon overwritten with `25.05` anways, when we publish the new release.

src/Particle/ParticleContainer.cpp
```diff
@@ -14,21 +14,66 @@ namespace
...
+    py::object pack_cpus (py::array_t<uint64_t> idcpus,
+                          py::array_t<int> cpus)
+    {
+        if (idcpus.ndim() != 1) {
+            throw std::runtime_error("Input should be 1-D NumPy array");
+        }
+
+        auto buf = idcpus.request();
+        auto buf2 = cpus.request();
+        if (buf.size != buf2.size) throw std::runtime_error("sizes do not match!");
```
EZoni: Shall we enclose this in parenthesis just for consistency with the other one-line if conditions?
```suggestion
        if (buf.size != buf2.size) {
            throw std::runtime_error("sizes do not match!");
        }
```

src/Particle/ParticleContainer.cpp
```diff
@@ -14,21 +14,66 @@ namespace
...
+    py::object pack_ids (py::array_t<uint64_t> idcpus,
+                         py::array_t<amrex::Long> ids)
+    {
+        if (idcpus.ndim() != 1) {
+            throw std::runtime_error("Input should be 1-D NumPy array");
+        }
+
+        auto buf = idcpus.request();
+        auto buf2 = ids.request();
+        if (buf.size != buf2.size) throw std::runtime_error("sizes do not match!");
```
EZoni: Shall we enclose this in parenthesis just for consistency with the other one-line `if` conditions?
```suggestion
        if (buf.size != buf2.size) {
            throw std::runtime_error("sizes do not match!");
        }
```

tests/test_particleTile.py
```diff
@@ -156,8 +156,55 @@ def test_ptile_aos_3d():
...
+    idcpu = np.array(
+        [
+            9223372036871553124,
+            9223372036871553124,
+            9223372036871553124,
+            9223372036871553124,
+            9223372036871553124,
+        ],
+        dtype=np.uint64,
+    )
```
EZoni, line 168: Maybe we could add an inline comment clarifying where this number comes from, just to make the test as clear as possible.

EZoni (commented):
Thanks, @atmyers! I left a few minor comments, please consider if the suggestions are worth being applied. After this round of review the PR is ready to merge, as far as I'm concerned. I think it would be nice to have this feature before we release 25.05.

EZoni (approved):
Thanks a lot!

PR #436, opened 2025-04-16
Bump Ubuntu from 20.04 -> 24.04 in CI

atmyers:

PR #438, opened 2025-04-23
For MuliFab indexing, () means all cells in all dimension, valid and ghost

dpgrote:
This PR clarifies the use of the empty tuple index, `()`, for MultiFabs. Giving the empty tuple as the only index of a MultiFab will return the full array, both valid and ghost cells. This is in contrast to the `Ellipsis` that will return the array with only value cells.
For example
```
mf = MultiFab()
mf[()]  # returns valid and ghost cells
mf[...]  # returns only valid cells
```

### DIFF
--- a/src/amrex/extensions/MultiFab.py
+++ b/src/amrex/extensions/MultiFab.py
@@ -316,9 +316,9 @@ def _process_index(self, index):
         # If only one slice or integer passed in, it was not wrapped in a tuple
         index = [index]
     elif isinstance(index, tuple):
-        if len(index) == 0:
-            # The empty tuple specifies all valid and ghost cells
-            index = [index]
+        if len(index) == 0: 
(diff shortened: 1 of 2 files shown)


roelof-groenewald (approved):
Thanks for the update!

EZoni:
Note that #436 (plus resetting the required workflows in the repo's settings, if any workflow's name changed) may be needed here to unblock CI.

roelof-groenewald:
@EZoni is it okay if I assign you to this PR then since you are familiar with how to reset the necessary repo settings?

EZoni:
@roelof-groenewald 

Thanks, you can assign me so I can keep track of this and make sure it does not get stalled, but unfortunately I do not have access to the settings of this repository (nor those of WarpX, FYI). I should be able to touch base tomorrow with @atmyers on #436 to see how to progress there.

PR #439, opened 2025-04-23
Pywrapper for ba convert, surroundingnodes, and mf.copy

RevathiJambunathan:


src/Base/MultiFab.H
```diff
@@ -493,6 +493,31 @@ factory :
         ;
     }
 
+    py_MultiFab
+        .def("copymf",
```
RevathiJambunathan, line 497: for some reason, calling this `copy` did not seem to work 
May be there is another copy method that is shadowing this 
so I renamed it as `copymf` and that works!

atmyers: Is this different than the copy method exposed [here](https://github.com/AMReX-Codes/pyamrex/blob/development/src/Base/MultiFab.cpp#L106)?

RevathiJambunathan: I tried using the copy_mfab in the python script and it did not work 
So now I am not sure 
they look similar, but `copy_mfab` did not work for me

atmyers: What if you use the extension method [here](https://github.com/AMReX-Codes/pyamrex/blob/development/src/amrex/extensions/MultiFab.py#L654), as in [this](https://github.com/AMReX-Codes/pyamrex/blob/development/tests/test_multifab.py#L492) test?

atmyers (approved):

src/Base/MultiFab.H
RevathiJambunathan: discussed offline 
.`copy` creates another copy and does not do what we want in this PR
where we copy the values from a src to dst multifab and the dst multifab is already allocated and initialized.

PR #440, opened 2025-05-02
Fixes in MultiFab setitem

dpgrote:
Previously, `setitem` was only checking for numpy arrays, which was causing errors on a GPU when a cupy array was passed in. This fix makes it more flexible, so now lists can also be used for example.

Also, when using GPU, the input value must be converted to a cupy array before the copy can be done.

A minor cleanup of `_get_field` is also done, simplifying the code.


RevathiJambunathan:
Thank you @dpgrote for this PR!!

atmyers (approved):

PR #441, opened 2025-05-02
Fix HIP CI by disabling IPO

WeiqunZhang:


.github/workflows/hip.yml
EZoni, line 1: Thanks for looking more into the issue, @WeiqunZhang!

Shall we use the same HIP version that we use in WarpX's CI, for homogeneity? Just a suggestion, it's fine with me to keep things as is.

For reference, here's what we did in WarpX:
- https://github.com/BLAST-WarpX/warpx/pull/5707
- https://github.com/BLAST-WarpX/warpx/pull/5731

WeiqunZhang: The default is the same I think.

EZoni: Oh yes, that's right. Sorry, somehow I remembered it was 6.4.2. My bad, all good then.

.github/workflows/hip.yml
```diff
@@ -46,20 +46,12 @@ jobs:
...
             -DCMAKE_VERBOSE_MAKEFILE=ON  \
+            -DpyAMReX_IPO=OFF            \
             -DAMReX_GPU_BACKEND=HIP      \
-            -DAMReX_SPACEDIM="1;2;3"     \
-            -DCMAKE_C_COMPILER=$(which clang)             \
-            -DCMAKE_CXX_COMPILER=$(which clang++)         \
-            -DAMReX_AMD_ARCH=gfx1010                      \
-            -DAMReX_ROCTX=ON                              \
-            -DCMAKE_CXX_STANDARD=17
+            -DAMReX_AMD_ARCH=gfx90a      \
```
EZoni, line 53: In WarpX, https://github.com/BLAST-WarpX/warpx/blob/development/.github/workflows/hip.yml, we seem to have `-DAMReX_AMD_ARCH=gfx900` instead of `-DAMReX_AMD_ARCH=gfx90a`. Do you happen to know why and/or if that matters at all?

WeiqunZhang: I don't think it matters at all. gfx90a is what's on frontier. gfx900 is quite old.

WeiqunZhang: I know for sure is that gfx90a with IPO does not work.

EZoni (approved):

atmyers (approved):

PR #445, opened 2025-05-12
Fix RealVect crossProduct

AlexanderSinn:
Another fix for https://github.com/AMReX-Codes/amrex/pull/4430

### DIFF
--- a/.github/workflows/hip.yml
+++ b/.github/workflows/hip.yml
@@ -51,8 +51,7 @@ jobs:
             -DpyAMReX_IPO=OFF            \
             -DAMReX_GPU_BACKEND=HIP      \
             -DAMReX_AMD_ARCH=gfx90a      \
-            -DAMReX_SPACEDIM="1;2;3"     \
-            -DpyAMReX_amrex_branch="development"
+            -DAMReX_SPACEDIM="1;2;3"
         cmake --build build --target pip_install -j 4
 
         ccache -s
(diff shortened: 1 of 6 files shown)


AlexanderSinn:
I think with `.def("crossProduct", &RealVect::crossProduct<>,` it also worked with amrex development but not amrex release. I only converted about half the CI tests to development. But with `pyamrex_realvect_crossproduct` it works with both versions.

WeiqunZhang:
Oh I didn't realize that the failed tests were still using an old version
of amrex. No wonder those error messages didn't make sense. Anyway, please
revert my changes because it is too complicated.

WeiqunZhang (approved):

WeiqunZhang:
Should we switch to using amrex development in CI?

AlexanderSinn:
Probably should. However, I don't know how to change the amrex branch in the tests where pyamrex is compiled with pip

WeiqunZhang:
That might be difficult.

ax3l:
@WeiqunZhang @AlexanderSinn I rather would, as in other projects, add pyAMReX to the app tests in AMReX and then manually update here the commits weekly (and on demand for breaking PRs). That gives us more control

ax3l:
Only `cmake/dependencies/AMReX.cmake` needs to be updated here, as we do in WarpX/ImpactX/HiPACE++.

See #446 and #379 as examples

PR #447, opened 2025-06-10
ParticleContainer: add_particle(other) & make_alike

ax3l:
Add bindings to copying from another particle container and to create another particle container from an existing one using the same layout & allocator.


ax3l:
@EZoni @atmyers does this look good to merge for you?

PR #454, opened 2025-07-23
Implement DLPack

### LINKED ISSUE #9: Discussion on mapping between amrex, numpy.ndarray, and torch.tensor data types

ax3l:
Add first-class support for zero-copy data exchange with ROCm and SYCL GPUs via DLPack interfaces.

Specs:
- https://dmlc.github.io/dlpack/latest/python_spec.html#implementation
- https://github.com/dmlc/dlpack/blob/v1.1/include/dlpack/dlpack.h

DLPack is implemented **once** in a shared `src/dlpack/DLPack.H`/`DLPack.cpp`: the full `__dlpack__` keyword protocol (`stream`/`max_version`/`dl_device`/`copy`), capsule creation, producer keep-alive, `copy=True` and device-to-host transfers, and the read-only flag. Each container binds it with a small `DLPackInfo` adapter, so there is no per-class duplication. `src/dlpack/dlpack.h` is re-vendored byte-identical from upstream v1.1, and we emit **both** the versioned (`"dltensor_versioned"`) and legacy (`"dltensor"`) capsules based on the consumer's `max_version` — so recent and older NumPy/CuPy/PyTorch/dpnp all work without forcing a version bump.

Like CuPy for CUDA/ROCm, dpnp on SYCL is an optional dependency that is imported lazily, only when actually used (`.to_dpnp()` / SYCL `.to_xp()`).

Tested on CPU, CUDA (RTX A2000), ROCm/HIP (MI250X) and SYCL (Intel Iris Xe, Level Zero) with numpy / cupy / dpnp / torch.

Close #9

## Action Items

- [x] start by vibing while preparing dinner, then manually:
- [x] review and finish Array4
- [x] PODVector
- [x] Vector
- [x] ~ArrayOfStructs~: skipped — record (struct) element types are not representable in DLPack; AoS keeps its structured `__array_interface__` / `__cuda_array_interface__`, and its per-component SoA data is DLPack-capable via PODVector
- [x] BaseFab
- [x] SmallMatrix (`long double` is compile-time skipped: no DLPack dtype)
- [x] SYCL: Implement `.to_dpnp` helper functions
- [x] Update `.to_xp` functions to use `.to_dpnp` (from [dpctl](https://github.com/IntelPython/dpctl)) for SYCL GPUs
- [x] Test on CUDA GPU: RTX A2000
- [x] Test on ROCm GPU: MI250X on Frontier (OLCF)
- [x] Test on SYCL GPU: Intel Iris Xe
- [x] Search docs for needed updates.
- [x] Fix DLPack stubs ~https://github.com/sizmailov/pybind11-stubgen/pull/258 or bind manually in pyAMReX~ — signatures are stubgen-clean as written (kwargs typed `typing.Any`, return `types.CapsuleType`); the CI stub job now pins upstream `pybind/pybind11-stubgen`

## Follow-ups (out of scope for this PR)

- **Exit-time SYCL teardown segfault** (cosmetic): once a dpnp array has been imported from an AMReX *device* pointer via DLPack, the bare interpreter can segfault at process exit *after* `amrex.finalize()` — dpctl and AMReX each own a `sycl::context` over the same device and tear them down in a bad order. It does not affect correctness or the test suite (`os._exit(0)` after finalize is clean). The proper fix is AMReX-side: have `Device::Initialize` adopt an existing external SYCL context instead of always creating its own, mirroring the `MPI_Initialized` → `call_mpi_finalize = 0` guard in `ParallelDescriptor`.


roelof-groenewald:
I performed some testing of the new functionality on Perlmutter. After the latest commit, the following appears to work as intended:
```
def test_mfab_cuda_cupy(mfab_device):
    import cupy as cp

    # AMReX -> cupy
    for mfi in mfab_device:   
        marr_cupy_from_dlpack = cp.from_dlpack(mfab_device.array(mfi))
        marr_cupy_from_dlpack[0, 1, 3, 2] = 5

    for mfi in mfab_device:   
        marr_cupy_from_dlpack = cp.from_dlpack(mfab_device.array(mfi))
        print(marr_cupy_from_dlpack[0, 1, 3, 2])
```
It executes without failure and prints the modified value `5`. Inspection of the  `DLDevice` showed that the device was successfully identified as `kDLCUDA`. The device id returned `3`, which is consistent with Perlmutter's standard rank-to-gpu mapping with just one MPI rank.

ax3l:
Awesome, then we are nearly there.

Try the dpnp logic for SYCL next?

roelof-groenewald:
I tested the `dlpack` functionality on Aurora (SYCL) and it also now produces the expected result. I also modified `Array4_to_xp` to take into account the GPU backend. We can now successfully access a MultiFab's `Array4` from a SYCL device with
```
for mfi in mfab_device:
    mfab_device.array(mfi).to_dpnp()
```

src/Base/Array4.H
```diff
@@ -220,15 +222,79 @@ namespace pyAMReX
...
             // https://docs.cupy.dev/en/stable/user_guide/interoperability.html#dlpack-data-exchange-protocol
+            .def("__dlpack__", [](
+                    Array4<T> const &a4,
+                    py::kwargs kwargs
+                    /* TODO: Handle keyword arguments
+                    [[maybe_unused]] py::handle stream,
+                    [[maybe_unused]] std::tuple<int, int> max_version,
+                    [[maybe_unused]] std::tuple<DLDeviceType, int32_t> dl_device,
+                    [[maybe_unused]] bool copy
+                    */
```
roelof-groenewald: Just want to flag this since `copy=True` doesn't yet work in the `.to_dpnp()` function.

roelof-groenewald:
I compiled WarpX on Aurora using this pyamrex branch. With it I was able to successfully run a multi-GPU simulation that uses `fields.py` to read MultiFab values  🎉 🚀

src/amrex/extensions/MultiFab.py
```diff
@@ -132,15 +172,14 @@ def mf_to_xp(self, copy=False, order="F"):
...
-    import inspect
+    views = []
+    for mfi in self:
+        views.append(self.array(mfi).to_xp(copy, order))
 
-    amr = inspect.getmodule(self)
-    return (
-        self.to_cupy(copy, order) if amr.Config.have_gpu else self.to_numpy(copy, order)
-    )
+    return views
```
roelof-groenewald, line 102: By changing the `mf_to_xp` function to directly call `array4_to_xp` (instead of doing the same configuration check as in `array4_to_xp` and then calling individual `mf_to_numpy/cupy/dpnp`) we no longer need the `mf_to_cupy` or `mf_to_dpnp` functions. I imagine we would want to keep `mf_to_numpy` in case someone has a specific need to return a numpy array during GPU execution, but I don't see the need to have the `cupy` or `dpnp` versions. @ax3l do you agree?

ax3l: Agreed, that looks sensible to me.

ax3l: On 2nd thought though, the docs mention the explicit functions and might need a rewrite when we remove to_cupy et al. Having those available is intuitive.

But always using to_xp and to_numpy is the right way to write portable code...

ax3l:
We need to rebase against `development` after #455 was merged. I already added the `DLDeviceType` bindings now and the other PR adds capsule type hints.

src/dlpack/DLPack.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+
+void init_DLPack(py::module& m)
+{
+    using namespace amrex;
+
+    // register types only if not already present, e.g., from another library
+    // that also implements DLPack bindings and exposes the types
+
+    py::type pyDLDeviceType = py::type::of<DLDeviceType>();
```
ax3l:
```suggestion
    py::type pyDLDeviceType = false;  // TODO: py::type::of<DLDeviceType>();
```

src/dlpack/DLPack.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+
+void init_DLPack(py::module& m)
+{
+    using namespace amrex;
+
+    // register types only if not already present, e.g., from another library
+    // that also implements DLPack bindings and exposes the types
+
+    py::type pyDLDeviceType = false;  // TODO: py::type::of<DLDeviceType>();
```
ax3l:
```suggestion
    // TODO: py::type pyDLDeviceType = py::type::of<DLDeviceType>();
    bool pyDLDeviceType = false;
```

src/dlpack/DLPack.cpp
```diff
@@ -0,0 +1,35 @@
...
+            .value("kDLMetal", DLDeviceType::kDLMetal)
+            .value("kDLVPI", DLDeviceType::kDLVPI)
+            .value("kDLROCM", DLDeviceType::kDLROCM)
+            .value("kDLROCMHost", DLDeviceType::kDLROCMHost)
+            .value("kDLExtDev", DLDeviceType::kDLExtDev)
+            .value("kDLCUDAManaged", DLDeviceType::kDLCUDAManaged)
+            .value("kDLOneAPI", DLDeviceType::kDLOneAPI)
+            .value("kDLWebGPU", DLDeviceType::kDLWebGPU)
+            .value("kDLHexagon", DLDeviceType::kDLHexagon)
+            .value("kDLMAIA", DLDeviceType::kDLMAIA)
```
ax3l, line 31:
```suggestion
            .value("kDLMAIA", DLDeviceType::kDLMAIA)
            .export_values()
            .finalize()
```

docs/source/install/dependencies.rst
```diff
@@ -29,7 +30,8 @@ Optional dependencies include:
 - `Python dependencies <https://www.python.org>`__
 
   - `mpi4py 2.1+ <https://mpi4py.readthedocs.io>`__: for multi-node and/or multi-GPU execution
-  - `cupy 12.2+ <https://github.com/cupy/cupy#installation>`__
+  - `cupy 12.2+ <https://github.com/cupy/cupy#installation>`__ (13.0+ for DLPack copy/device requests): for CUDA/ROCm GPU zero-copy views
```
ax3l:
```suggestion
  - `cupy 13.0+ <https://github.com/cupy/cupy#installation>`__: for CUDA/ROCm GPU zero-copy views
```

docs/source/usage/zerocopy.rst
```diff
@@ -57,6 +69,18 @@ AI/ML: pyTorch
...
+
+Everything Else: DLPack
+-----------------------
+
+Frameworks not covered above can exchange data with pyAMReX through the standardized `DLPack <https://dmlc.github.io/dlpack/latest/python_spec.html>`__ protocol, on CPU as well as on CUDA, ROCm and SYCL GPUs.
+
+Data objects of pyAMReX implement ``__dlpack__`` and ``__dlpack_device__`` and can be consumed by, e.g., ``numpy.from_dlpack``, ``cupy.from_dlpack``, ``dpnp.from_dlpack``, ``torch.from_dlpack`` or ``jax.dlpack``.
+Both DLPack 1.x ("versioned") and legacy consumers are supported, including device-to-host transfers via ``from_dlpack(..., device="cpu")``.
+
+An exception is the particle ``ArrayOfStructs``: DLPack cannot describe its record (struct) element type.
```
ax3l:
```suggestion
An exception is the `deprecated <https://github.com/AMReX-Codes/pyamrex/issues/459>`__ particle ``ArrayOfStructs``: DLPack cannot describe its record (struct) element type.
```

src/amrex/extensions/dlpack_helpers.py
```diff
@@ -0,0 +1,124 @@
...
+    """Import a pyAMReX object into CuPy via DLPack.
+
+    Device data is imported zero-copy (or as an isolated device-side copy
+    for ``copy=True``). Host-side data is always copied to the device,
+    since a cross-device view is not possible.
+    """
+    import cupy as cp
+
+    device_type, _ = self.__dlpack_device__()
+    if device_type == kDLCPU:
```
ax3l: Codex:

This branch only treats `kDLCPU` as host memory. CUDA/ROCm pinned allocations are advertised as `kDLCUDAHost`/`kDLROCMHost`, so they fall through to `cp.from_dlpack`, which CuPy rejects. I reproduced this with `HostVector_real.to_cupy()`: it raises `TypeError: Unsupported array type: 3`, while the pre-PR `cp.array(..., copy=False)` path succeeds. Please route all three host-accessible device types through the host-to-device copy branch.

src/dlpack/DLPackHelpers.cpp
```diff
@@ -0,0 +1,539 @@
...
+            } else {
+                to_host = true;
+            }
+        }
+
+        bool const want_copy =
+            to_host || (copy_request.has_value() && copy_request.value());
+
+        std::function<void()> free_data;
+        if (want_copy) {
```
ax3l, line 603: Codex:

The copy branch bypasses `handle_stream()` entirely. As a result, `arr.__dlpack__(stream=1, copy=True)` succeeds for a CPU tensor even though DLPack permits only `None` there; CUDA/ROCm copies likewise skip stream type/value validation and run on the AMReX producer stream. The protocol also requires a copy to use the consumer stream when it is understood. Please validate `stream` on both paths and either perform the copy over the requested stream or reject unsupported stream/copy combinations.

src/Base/PODVector.cpp
```diff
@@ -350,6 +353,43 @@ void add_host_device(PODVectorClass &... cls)
...
+{
+    namespace dlpack = pyAMReX::dlpack;
+
+    if constexpr (dlpack::has_dlpack_dtype_v<T>)
+    {
+        dlpack::add_dlpack(cl, [](amrex::PODVector<T, Allocator> const & podvector) {
+            dlpack::DLPackInfo info;
+            info.data = const_cast<void*>(
+                static_cast<void const*>(podvector.dataPtr()));
+            info.device = dlpack::detect_device_from_pointer(podvector.dataPtr());
```
ax3l: Codex:

Inferring the device solely from `dataPtr()` misclassifies newly empty GPU vectors because their pointer is null. On the CUDA build I observed `NonManagedDeviceVector_real` report `(kDLCPU, 0)` while empty and `(kDLCUDA, 0)` after one `push_back`; managed and pinned vectors change similarly. That gives `__dlpack_device__()` and direct consumers the wrong logical device for empty tensors. Please classify an empty PODVector from its allocator/arena instead.

src/dlpack/DLPack.cpp
```diff
@@ -0,0 +1,539 @@
...
+        DLDevice device {kDLCPU, 0};
+
+#if defined(AMREX_USE_CUDA)
+        cudaPointerAttributes attr;
+        cudaError_t const err = cudaPointerGetAttributes(&attr, ptr);
+        if (err == cudaSuccess) {
+            if (attr.type == cudaMemoryTypeDevice) {
+                device = DLDevice{kDLCUDA, attr.device};
+            } else if (attr.type == cudaMemoryTypeManaged) {
+                device = DLDevice{kDLCUDAManaged, attr.device};
```
ax3l: Codex:

DLPack requires `device_id` to be zero for managed memory (the vendored `dlpack.h` states this as well). Exporting `attr.device` produces a nonzero ID for processes assigned CUDA device 1 or higher and can confuse consumers that follow the managed-memory convention. Please export `DLDevice{kDLCUDAManaged, 0}` here.

ax3l:
@roelof-groenewald crazy enough, with 3 LLMs involved, this now works well on CPU, CUDA, ROCm, and SYCL.

roelof-groenewald:
> @roelof-groenewald crazy enough, with 3 LLMs involved, this now works well on CPU, CUDA, ROCm, and SYCL.

Awesome!! Is this ready to be merged then?

ax3l:
Yep, this ready!

ROCm support is now seamless again (since cupy dropped `__cuda_array_interface__` in newer versions).
SYCL now works up to a teardown issue on process end, which we will need to address in follow-up PRs to both AMReX and pyAMReX.

roelof-groenewald (approved):

PR #457, opened 2025-07-29
Doxygen: No Python Files

ax3l:
You are too eager, Doxygen. We want you to just read C++.


ax3l:
This is not an issue yet, we just saw it causing issues in WarpX. https://github.com/BLAST-WarpX/warpx/pull/6052

PR #458, opened 2025-07-31
Single Precision Support

ax3l:
- [x] `setup.py`
- [x] CI coverage
- [x] fix runtime issues in tests

As a follow-up, one could add support for SP `.to_df()` by removing binary columns (members) from the AoS before passing it into Pandas. But this legacy layout is not our need or focus, nor worth spending more time on.

### DIFF
--- a/src/Particle/ArrayOfStructs.H
+++ b/src/Particle/ArrayOfStructs.H
@@ -43,16 +43,37 @@ namespace
 #if (AMREX_SPACEDIM >= 3)
         descr.append(py::make_tuple("z", py::format_descriptor<RealType>::format()));
 #endif
+
+        /* Note AMReX_Particle.H:
+         * struct alignas(sizeof(double)) Particle ...
+         *   aligns the ParticleBase member (pads the end if needed).
+         * in
+         * struct ParticleBase
+         * {
+         *   T m_pos[AMREX_SPACEDIM];
+         *   T m_rdata[NReal];
+         *   uint64_t m_idcpu = 0;
+         *   int m_idata[NInt];
+         * };
+         */
         if constexpr (ParticleType::NReal > 0) {
             for(int ii=0; ii < ParticleType::NReal; ii++) {
                 descr.append(py::make_tuple("rdata_"+std::to_string(ii),py::format_descriptor<RealType>::format()));
             }
         }
+#ifdef AMREX_SINGLE_PRECISION_PARTICLES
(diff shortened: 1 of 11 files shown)


tests/test_aos.py
```diff
@@ -76,15 +76,16 @@ def test_array_interface():
     # print('particle 2 from aos:\n',aos[1])
     # print('array interface\n', aos.__array_interface__)
     arr = aos.to_numpy()
+    int_arg = 7 if arr[0][0].dtype == "float32" else 6  # padding
```
dpgrote, line 79: Is this something that users will have to know, that there might be this extra padding element so the indices for the integers might be shifted?

ax3l: I fear so... I have not found a way to kick out numpy struct elements (in the SP case: the paddings) without doing a copy yet.

That said, nobody should use AoS anymore in AMReX. I am inclined to drop all support for it in pyAMReX to stop wasting time :)

src/Particle/ArrayOfStructs.H
```diff
@@ -43,17 +43,38 @@ namespace
...
         descr.append(py::make_tuple("idcpu", py::format_descriptor<uint64_t>::format()) );
         if constexpr (ParticleType::NInt > 0) {
             for(int ii=0; ii < ParticleType::NInt; ++ii) {
                 descr.append(py::make_tuple("idata_"+std::to_string(ii),py::format_descriptor<int>::format()));
             }
-        }
+        }/*
+        if constexpr (ParticleType::NInt % 2 != 0) {  // alignas
+            descr.append(py::make_tuple("", "|V4"));  // empty space due to alignment
+        }*/
```
ax3l, line 77: Adding this end-of-particle padding after the `int` (32bit) array fails both the double precision tests o.0

tests/test_aos.py
ax3l: Solicited feedback now on #459 and on Slack

ax3l:
There seems to be a follow-up issue in SP in ImpactX (which uses pure SoA particle layout) for `.to_df()`:
https://github.com/conda-forge/impactx-feedstock/pull/56#issuecomment-3304382067
Issue is a segfault with `SoA_np` in https://github.com/AMReX-Codes/pyamrex/blob/25.09/src/amrex/extensions/StructOfArrays.py#L41 https://github.com/AMReX-Codes/pyamrex/blob/25.09/src/amrex/extensions/PODVector.py#L35

Note the PR here just works around AoS stuff (legacy).

PR #474, opened 2025-09-09
Implement same dependency/release update workflow as WarpX

EZoni:
Requested by @ax3l on Slack (to be used in #378):

> Do you mind adding the Tools/Release/update_dependencies.py script (simply in /) to pyAMReX, too? Do not need the other automations for now.

WarpX references:
- https://github.com/BLAST-WarpX/warpx/pull/5965
- https://github.com/BLAST-WarpX/warpx/pull/5993
- https://github.com/BLAST-WarpX/warpx/pull/6038

To-do:
- [x] AMReX
- [x] pybind11


update_dependencies.py
EZoni, line 1: Down the road it would be nice if we found a way to make this a BLAST-level module, since a lot of code is shared with the WarpX script.

ax3l (changes requested):
I think we need to include `dependencies.json` in `MANIFEST.in`, otherwise `PyPI` source packages (uncompiled) will lack the file and CMake builds from pip source packages will fail.

ax3l:
I'll add it and post a fix to WarpX, too

ax3l (approved):
Thank you! :)

PR #477, opened 2025-09-16
Fix MultiFab getitem for single imaginary index

dpgrote:
This fixes the indexing of MultiFabs when a single imaginary index is given, for example `Ex[1j]`.


roelof-groenewald (approved):
Thanks!

PR #478, opened 2025-09-17
Fix Use-After-Free in NumPy/df

ax3l:
For AoS and SoA `.to_numpy()` and dependent logic
like `.to_df()`, the lifetime of the temporary copied variable is only handled by NumPy if we first create a named variable.

The `to_host()` returned object is a temporary, and `np.array` using the `__array_interface__` protocol does not keep it alive automatically unless it is stored in an actual variable (eg., `tmp`).

X-ref:
- https://github.com/conda-forge/impactx-feedstock/pull/56#issuecomment-3304382067
- https://github.com/AMReX-Codes/pyamrex/pull/458#issuecomment-3304399453

### DIFF
--- a/src/amrex/extensions/PODVector.py
+++ b/src/amrex/extensions/PODVector.py
@@ -29,10 +29,14 @@ def podvector_to_numpy(self, copy=False):
         if copy:
             # This supports a device-to-host copy.
             #
-            # todo: validate of the to_host() returned object
-            #       lifetime is always managed correctly by
-            #       Python's GC - otherwise copy twice via copy=True
-            return np.array(self.to_host(), copy=False)
+            # The to_host() returned object is a temporary, and
+            # np.array using the __array_interface__ protocol does
+            # not keep it alive automatically unless it is stored
+            # in an actual variable (tmp).
+            tmp = self.to_host()
+            ret = np.array(tmp, copy=False)
+            assert ret.base is tmp
+            return ret
         else:
             return np.array(self, copy=False)
     else:
--- a/src/amrex/extensions/ArrayOfStructs.py
+++ b/src/amrex/extensions/ArrayOfStructs.py
@@ -32,10 +32,14 @@ def aos_to_numpy(self, copy=False):
     if copy:
         # This supports a device-to-host copy.
         #
-        # todo: validate of the to_host() returned object
-        #       lifetime is always managed correctly by
-        #       Python's GC - otherwise copy twice via copy=True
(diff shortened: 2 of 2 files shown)


src/amrex/extensions/PODVector.py
```diff
@@ -29,10 +29,14 @@ def podvector_to_numpy(self, copy=False):
         if copy:
```
ax3l, line 29: We can generally make this a bit more explicit by checking if `self` is on GPU and otherwise just copying from it directly, with another `if`/`else` branch. But the logic here should have the same effect and amount of copies involved...


The only difference is that here we will always end up with pinned memory, while we might be sometimes interested using regular host memory. But seems niche.

roelof-groenewald (approved):
Looks good to me.
Just one question: Do we need the `assert`s to "trick" Python into keeping the object alive?

ax3l:
The assert is not needed, but if it is ever violated I want a python error instead of a segfault in code that uses the returned variable. I was too lasy to `raise`, but maybe that would be cleaner?

ax3l:
Ah, `assert` just raises `AssertionError` in Python, so its good enough for here.

src/amrex/extensions/PODVector.py
```diff
@@ -29,10 +29,14 @@ def podvector_to_numpy(self, copy=False):
...
             #
-            # todo: validate of the to_host() returned object
-            #       lifetime is always managed correctly by
-            #       Python's GC - otherwise copy twice via copy=True
-            return np.array(self.to_host(), copy=False)
+            # The to_host() returned object is a temporary, and
+            # np.array using the __array_interface__ protocol does
+            # not keep it alive automatically unless it is stored
+            # in an actual variable (tmp).
+            tmp = self.to_host()
```
ax3l, line 36: Hm, looks like there is more / this is not a sufficient fix (at least for some SP tests on conda-forge):
https://github.com/conda-forge/impactx-feedstock/pull/56#issuecomment-3308240587

It already has an issue here, so the problem could be in the implementation of `to_host()`...
https://github.com/AMReX-Codes/pyamrex/blob/25.09/src/Base/PODVector.cpp#L79-L87

ax3l: Very hard to reproduce locally... also cannot spot anything in valgrind yet.

ax3l: seems to be fixed by https://github.com/BLAST-ImpactX/impactx/pull/1156 o.0

PR #482, opened 2025-09-23
Fix: `Config.precision`

ax3l:
Only `AMREX_USE_FLOAT` exists...

X-ref https://github.com/conda-forge/impactx-feedstock/pull/56#issuecomment-3323003288

### DIFF
--- a/src/Base/AMReX.cpp
+++ b/src/Base/AMReX.cpp
@@ -102,7 +102,7 @@ void init_AMReX(py::module& m)
         .def_property_readonly_static(
             "precision",
             [](py::object){
-#ifdef AMREX_SINGLE_PRECISION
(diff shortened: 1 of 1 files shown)


src/Base/AMReX.cpp
```diff
@@ -102,7 +102,7 @@ void init_AMReX(py::module& m)
         .def_property_readonly_static(
             "precision",
             [](py::object){
-#ifdef AMREX_SINGLE_PRECISION
+#ifdef AMREX_USE_FLOAT
```
ax3l, line 105: Still not sure this is correct...

ax3l: All good.

EZoni (approved):

PR #486, opened 2025-10-02
Strip "v" prefix from version for find_package

EZoni:
Somehow this fails in the builds for the conda-forge feedstock with
```
CMake Error at cmake/dependencies/pybind11.cmake:44 (find_package):
    find_package called with invalid argument "v3.0.1"
```

However, it doesn't fail in local builds, which is why I didn't catch it before the 25.10 release.

I will need to add this as a patch to https://github.com/conda-forge/pyamrex-feedstock/pull/52.


EZoni:
@WeiqunZhang 

Do you think this will work if we add it as a patch to https://github.com/conda-forge/pyamrex-feedstock/pull/52?

WeiqunZhang:
I think it will.

WeiqunZhang (approved):

PR #491, opened 2025-10-16
Update ParmParse

WeiqunZhang:


EZoni (approved):

EZoni:
Follow up to https://github.com/AMReX-Codes/amrex/pull/4695, if I'm not mistaken.

EZoni:
@WeiqunZhang 

I see that the pyAMReX CI check in https://github.com/AMReX-Codes/amrex/pull/4718 failed: https://github.com/AMReX-Codes/amrex/actions/runs/18621370224/job/53092959046.

Do we need another fix (which we could sneak into this PR)?

WeiqunZhang:
I don't think we need another fix. It's the same issue.

ax3l:
Oh, very cool. Thanks for the API update!

PR #493, opened 2025-10-23
CMake: Skip on Win `GNUInstallDirs`

ax3l:
Checking if this causes issues on Conda-Forge for Windows now... Somehow
```
GNUInstallDirs_get_absolute_install_dir(CMAKE_INSTALL_FULL_LIBDIR CMAKE_INSTALL_LIBDIR LIBDIR )
```
might expand our relative `lib` for `CMAKE_INSTALL_LIBDIR` to `/<SRC>/lib`...?


ax3l:
Does not fix the issue on CF

PR #497, opened 2025-11-03
Particle Iterator Shortcuts

ax3l:
* Allow to iterate on particles of level N without a level loop.
* Allow to access named pure SoA attributes directly on the particle iterator (`pti`).

[New simple interface example](https://pyamrex--497.org.readthedocs.build/en/497/usage/compute.html#particles): 
image

### DIFF
--- a/src/Particle/StructOfArrays.H
+++ b/src/Particle/StructOfArrays.H
@@ -57,11 +57,11 @@ void make_StructOfArrays(py::module &m, std::string allocstr)
              "Get access to a particle Real component Array (compile-time and runtime component)")
         .def("get_real_data", py::overload_cast<std::string const &>(&SOAType::GetRealData),
              py::return_value_policy::reference_internal,
-             py::arg("index"),
+             py::arg("name"),
              "Get access to a particle Real component Array (compile-time and runtime component)")
         .def("get_int_data", py::overload_cast<std::string const &>(&SOAType::GetIntData),
              py::return_value_policy::reference_internal,
-             py::arg("index"),
+             py::arg("name"),
              "Get access to a particle Real component Array (compile-time and runtime component)")
 
         // names
--- a/src/Particle/StructOfArrays.H
+++ b/src/Particle/StructOfArrays.H
@@ -55,6 +55,14 @@ void make_StructOfArrays(py::module &m, std::string allocstr)
              py::return_value_policy::reference_internal,
              py::arg("index"),
              "Get access to a particle Real component Array (compile-time and runtime component)")
+        .def("get_real_data", py::overload_cast<std::string const &>(&SOAType::GetRealData),
+             py::return_value_policy::reference_internal,
+             py::arg("index"),
+             "Get access to a particle Real component Array (compile-time and runtime component)")
+        .def("get_int_data", py::overload_cast<std::string const &>(&SOAType::GetIntData),
+             py::return_value_policy::reference_internal,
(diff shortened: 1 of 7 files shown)


src/amrex/extensions/ParticleContainer.py
```diff
@@ -6,7 +6,65 @@
...
+    if level == "all":
+        raise ValueError("level='all' is not yet supported for ParticleContainers")
+        # TODO: This does not work
+        # for lvl in range(self.finest_level + 1):
+        #     yield self.Iterator(self, level=lvl)
+    elif isinstance(level, int) and level >= 0:
+        return self.Iterator(self, level=level)
+    else:
+        raise ValueError(
+            f"level must be an integer in [0:finest_level+1) or 'all', but got: {level}"
```
roelof-groenewald:
```suggestion
            f"level must be an integer in [0:{self.finest_level+1}) or 'all', but got: {level}"
```

src/Particle/StructOfArrays.H
```diff
@@ -55,6 +55,14 @@ void make_StructOfArrays(py::module &m, std::string allocstr)
              py::return_value_policy::reference_internal,
              py::arg("index"),
              "Get access to a particle Real component Array (compile-time and runtime component)")
+        .def("get_real_data", py::overload_cast<std::string const &>(&SOAType::GetRealData),
+             py::return_value_policy::reference_internal,
+             py::arg("index"),
+             "Get access to a particle Real component Array (compile-time and runtime component)")
+        .def("get_int_data", py::overload_cast<std::string const &>(&SOAType::GetIntData),
+             py::return_value_policy::reference_internal,
+             py::arg("index"),
```
roelof-groenewald: Should we use `py::arg("attribute")` or something like that here instead?

roelof-groenewald (approved):
Looks good! Just two minor comments.

src/Particle/StructOfArrays.H
ax3l: Oh, wrong name. Maybe we should call this `name` as in the C++ api

ax3l:
```suggestion
             py::arg("name"),
             "Get access to a particle Real component Array (compile-time and runtime component)")
        .def("get_int_data", py::overload_cast<std::string const &>(&SOAType::GetIntData),
             py::return_value_policy::reference_internal,
             py::arg("name"),
```

src/amrex/extensions/ParticleContainer.py
```diff
@@ -6,7 +6,65 @@
...
+def iterator(self, *args, level=None):
+    """Create an iterator over all particle tiles
+
+    Parameters
+    ----------
+    self : amrex.ParticleContainer_*
+        A ParticleContainer class in pyAMReX
+    args : deprecated positional argument
+    level : int | str, optional
+        The MR level. Allowed values are [0:finest_level+1) and "all".
```
ax3l:
```suggestion
        The MR level. Allowed values are [0:self.finest_level+1) and "all".
```

src/amrex/extensions/ParticleContainer.py
ax3l: Good idea

src/Particle/StructOfArrays.H
roelof-groenewald: Sounds good!

PR #504, opened 2025-11-06
Start EB

WeiqunZhang:
This is still work in progress. You can build EB and create EBFArrayBoxFactory objects from python.


src/EB/EBFabFactory.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+    m.def(
+        "makeEBFabFactory",
+        [] (Geometry const& geom, BoxArray const& ba, DistributionMapping const& dm,
+            Vector<int> const& ngrow, EBSupport support)
+        {
+            return makeEBFabFactory(geom, ba, dm, ngrow, support);
+        },
+        py::arg("geom"), py::arg("ba"), py::arg("dm"), py::arg("ngrow"),
+        py::arg("support").
```
ax3l:
```suggestion
        py::arg("support"),
```

docs/source/usage/api.rst
```diff
@@ -280,3 +280,19 @@ This is for the legacy, AoS + SoA particle containers only:
 .. autoclass:: amrex.space3d.Particle_2_1
    :members:
    :undoc-members:
+
+.. _usage-api-eb:
+
+Embedded Boundaries
+-------------------
+
+Embedded boundary (EB) support is still experimental. To build pyAMReX with
```
ax3l: Although we use EBs in WarpX very well with Python... Maybe one can skip the first sentence completely or rephrase to:
```suggestion
Embedded boundary (EB) support in pyAMReX is still minimal. To build pyAMReX with
```

We only need more EB bindings if we build and allocate EBs from pyAMReX. If we just interact with EB-enabled simulations, we are already quite good (e.g., `iMultiFab` et al. is what we need for WarpX with EB).

src/EB/EBFabFactory.cpp
ax3l: fixed.

docs/source/usage/api.rst
ax3l: Updated

ax3l (approved):
Thank you for starting this! :)

tests/test_eb.py
```diff
@@ -0,0 +1,14 @@
...
+import amrex.space3d as amr
+
+
+@pytest.mark.skipif(not amr.Config.have_eb, reason="Requires -DAMReX_EB=ON")
+def test_makeEBFab():
+    pass
+
+    # TODO:
+    # EB2_Build(...)
+    # makeEBFabFactory(...)
```
ax3l: @WeiqunZhang do you like to finalize your PR by adding a test that uses your new functionality, so we have test coverage?
```suggestion
    ... = EB2_Build(...)
    ... = makeEBFabFactory(...)
```

WeiqunZhang: Yes, I will do it.

src/EB/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+foreach(D IN LISTS AMReX_SPACEDIM)
+    if (NOT AMReX_EB)
+        return()
+    endif()
+    if (D EQUAL 1)
+        return()
+    endif()
+
```
ax3l: Ah I see it. `foreach` is not a function :)

ax3l (approved):
Good start, thank you! :sparkles:

docs/source/usage/api.rst
```diff
@@ -280,3 +280,19 @@ This is for the legacy, AoS + SoA particle containers only:
...
+Embedded Boundaries
+-------------------
+
+Embedded boundary (EB) support in pyAMReX is still minimal. To build pyAMReX with
+EB support, you need to add ``-DAMReX_EB=ON`` to CMake build options.
+
+.. autofunction:: amrex.space3d.EB2_Build
+
+.. autoclass:: amrex.space3d.EBFArrayBoxFactory
+   :members:
```
ax3l, line 295: Hm, gotta build locally but something broke the autodocs for all types
https://pyamrex.readthedocs.io/en/latest/usage/api.html

ax3l: I built locally and the issue is https://github.com/sizmailov/pybind11-stubgen/pull/275

`EBFArrayBoxFactory` depends on `FabFactory_FArrayBox` but pybind11-stubgen does falsly an alphabetical sorting. Hoping this PR gets accepted and unbreaks it, otherwise we need to fork pybind11-stubgen.

ax3l: #509

PR #505, opened 2025-11-07
Fix PODVector resize and reserve

AlexanderSinn:
Needed after https://github.com/AMReX-Codes/amrex/pull/4763


src/Base/PODVector.cpp
```diff
@@ -72,9 +72,12 @@ void make_PODVector(py::module &m, std::string typestr, std::string allocstr)
...
         .def("empty", &PODVector_type::empty)
-        .def("resize", py::overload_cast<std::size_t>(&PODVector_type::resize))
-        .def("resize", py::overload_cast<std::size_t, const T&>(&PODVector_type::resize))
-        .def("reserve", &PODVector_type::reserve)
+        .def("resize", [](PODVector_type & pv, std::size_t new_size){
+            pv.resize(new_size); })
+        .def("resize", [](PODVector_type & pv, std::size_t new_size, const T& init_val){
+            pv.resize(new_size, init_val); })
+        .def("reserve", [](PODVector_type & pv, std::size_t new_capacity){
+            pv.reserve(new_capacity); })
```
ax3l: Thank you!

Let's improve this a bit to add:
- enums of the new arguments
- expose the new signatures with
- named arguments with
- default values

Please let me know if you need help.

AlexanderSinn: Is there a way to expose enum struct to python or should it just be a string?

WeiqunZhang: I did something like this in another place. It seems to work.
```
// C++
    py::enum_<EBSupport>(m, "EBSupport")
        .value("basic", EBSupport::basic)
        .value("volume", EBSupport::volume)
        .value("full", EBSupport::full);
```

```
# python
factory = amr.makeEBFabFactory(geom, ba, dm, ng, amr.EBSupport.full)
```

ax3l: Yep, you can use the new Enum support that I use in #454 as well, pybind11 recently modernized it. Both either works.

ax3l (approved):

PR #517, opened 2025-11-26
Always use version tag in pybind11 dependency update

EZoni:
Always use the version tag for pybind11 as we don't want to track updates on a weekly basis.

Related PRs:
- #474
- #486

@ax3l Does this fix the issue that we discussed offline?


ax3l (approved):
Thank you! Sorry I overlooked this :)

PR #527, opened 2026-01-16
Update Array4

WeiqunZhang:
The innard of amrex::Array has changed in https://github.com/AMReX-Codes/amrex/pull/4868.


ax3l (approved):
Thank you for the update! :)

PR #528, opened 2026-01-17
Update Array4

WeiqunZhang:
The stride member of Array4 has changed in https://github.com/AMReX-Codes/amrex/pull/4900.

### DIFF
--- a/src/Base/Array4.H
+++ b/src/Base/Array4.H
@@ -176,7 +176,7 @@ namespace pyAMReX
                 // which we do not need to set. For py::buffer_info, 2 is
                 // the unit stride direction. If AMReX has a set function,
                 // we would do
-                //  a4.set_stride<0>(buf.strides.at(2)); // not needed.,
+                //  a4.set_stride<0>(buf.strides.at(2)); // not needed
                 //  a4.set_stride<1>(buf.strides.at(1));
                 //  a4.set_stride<2>(buf.strides.at(0));
                 // Unfortunately, amrex::Array4 does not have a public function for setting
--- a/src/Base/Array4.H
+++ b/src/Base/Array4.H
@@ -63,7 +63,8 @@ namespace pyAMReX
     {
         auto d = py::dict();
         // F->C index conversion here
-        // p[(i-begin[0])+(j-begin[1])*nstride[0]+(k-begin[2])*nstride[1]+n*nstride[2]];
+        // p[(i-begin[0])+(j-begin[1])*jstride+(k-begin[2])*kstride+n*nstride];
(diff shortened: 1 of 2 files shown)


src/Base/Array4.H
```diff
@@ -165,16 +166,18 @@ namespace pyAMReX
...
                 a4.get()->end[1] = (int)buf.shape.at(1);
                 a4.get()->end[2] = (int)buf.shape.at(0);
                 a4.get()->end[3] = 1;
                 // buffer protocol strides are in bytes, AMReX strides are elements
-                a4.get()->nstride[0] = (int)buf.strides.at(1) / sizeof(T); // fastest varying index
-                a4.get()->nstride[1] = (int)buf.strides.at(0) / sizeof(T);
+                // Unfortunately, amrex::Array4 does not have a public function for
+                // setting strides. Note that stride.a[0] is get_stride<1>().
+                a4.get()->stride.a[0] = (int)buf.strides.at(1) / sizeof(T); // fastest varying index
+                a4.get()->stride.a[1] = (int)buf.strides.at(0) / sizeof(T);
```
EZoni: So the index mapping between left (`stride.a[index]`) and right (`strides.at(index)`) here is $0 \rightarrow 1$ and $1 \rightarrow 0$ (instead of $0 \rightarrow 1$ and $1 \rightarrow 2$)? Could you summarize again the index mapping in the various dimensions, just for clarity?

WeiqunZhang: Maybe I will put it as comments in the code.

WeiqunZhang:
```
                // buffer protocol strides are in bytes, AMReX strides are 
                // elements.  In AMReX, 0 is the unit stride direction, 
                // which we do not need to set. For py::buffer_info, 2 is 
                // the unit stride direction. If AMReX has a set function, 
                // we would do 
                //  a4.set_stride<0>(buf.strides.at(2)); // not needed 
                //  a4.set_stride<1>(buf.strides.at(1)); 
                //  a4.set_stride<2>(buf.strides.at(0)); 
                // Unfortunately, amrex::Array4 does not have a public function for setting 
                // strides, and stride<n> is stored at stride.a[n-1]. 
                a4.get()->stride.a[0] = (int)buf.strides.at(1) / sizeof(T);
                a4.get()->stride.a[1] = (int)buf.strides.at(0) / sizeof(T);
                // 3D == no component: stride here should not matter 
                a4.get()->stride.a[2] = a4.get()->template get_stride<2>() * (int)buf.shape.at(0);
```
Is that clear now?

EZoni (approved):

src/Base/Array4.H
EZoni: It's clearer now with the comment, thanks!

PR #540, opened 2026-03-02
PODVector: Add `_default` Platform Aliases

ax3l:
Update: Renamed what was added here in PR #545.
What I actually meant with this PR is `amrex::Gpu::DeviceVector<amrex::Real>`. When Cuda is off, all these containers revert to `amrex::Gpu::PODVector`.
https://github.com/AMReX-Codes/amrex/blob/26.03/Src/Base/AMReX_GpuContainers.H#L23-L24
https://github.com/AMReX-Codes/amrex/blob/26.03/Src/Base/AMReX_GpuContainers.H#L79-L81

Register `PODVector_{real,int,uint64}_default` as aliases that resolve to `_std` on CPU builds and `_arena` on GPU builds, matching the C++ `Gpu::DeviceVector<T>` type per platform.
For other containers like PC & MF we an just use the default arena, but PODVector is implemented a bit special, relying on a `std::allocator` instead of the default arena on CPU.

This helps to write CPU/GPU agnostic interfaces, e.g., in https://github.com/BLAST-ImpactX/impactx/pull/1305

### DIFF
--- a/src/Base/PODVector.cpp
+++ b/src/Base/PODVector.cpp
@@ -43,7 +43,14 @@ void make_PODVector(py::module &m, std::string typestr, std::string allocstr)
     auto const podv_name = std::string("PODVector_").append(typestr)
                            .append("_").append(allocstr);
 
-    py::class_<PODVector_type>(m, podv_name.c_str())
+    auto const podv_doc = std::string(
+        "A plain-old-data (POD) vector of '")
+        .append(typestr)
+        .append("' elements with '")
+        .append(allocstr)
+        .append("' allocation.");
+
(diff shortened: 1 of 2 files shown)


ax3l:
What I actually meant with this PR is `amrex::Gpu::DeviceVector<amrex::Real>`. When Cuda is off, all these containers revert to `amrex::Gpu::PODVector`.
https://github.com/AMReX-Codes/amrex/blob/26.03/Src/Base/AMReX_GpuContainers.H#L23-L24
https://github.com/AMReX-Codes/amrex/blob/26.03/Src/Base/AMReX_GpuContainers.H#L79-L81

Maybe should rename them in a follow-up to make it clearer: #545

PR #543, opened 2026-03-03
Python 3.11+

ax3l:
Document that we currently support Python 3.11-3.14. Python 3.10 is EOL end of Oct and does not support the same level of typing that we want to use more across our SW stacks.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -259,7 +259,7 @@ setup(
     ext_modules=cxx_modules,
     cmdclass=cmdclass,
     zip_safe=False,
-    python_requires=">=3.8",  # left for CI, truly ">=3.10"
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -259,7 +259,7 @@ def build_extension(self, ext):
     ext_modules=cxx_modules,
     cmdclass=cmdclass,
     zip_safe=False,
-    python_requires=">=3.8",  # left for CI, truly ">=3.10"
+    python_requires=">=3.8",  # left for CI, truly ">=3.11"
```
roelof-groenewald, line 262: How long do you expect before fully bumping this up to 3.11?

PR #555, opened 2026-04-03
CI: Try New Stubgen Sorting

ax3l:
Try new stubgen sorting from https://github.com/pybind/pybind11-stubgen/pull/238


ax3l:
Works well! In `development` now

PR #545, opened 2026-03-04
`amrex::Gpu::DeviceVector` et al.

ax3l:
Follow-up to #540, renaming the `PODVector_{real,int,uint64}_default` to `DeviceVector_{real,int,uint64}` et al. Implement the aliases for `AMReX_GpuContainers.H`

- [x] resolve inconsistency mentioned in https://github.com/AMReX-Codes/amrex/issues/5123

PR #556, opened 2026-04-05
`PodVector_*.from_*` NumPy/CuPy

ax3l:
Helpers to copy data directly from `NumPy`/`CuPy` to simplify user code, e.g., to add particles in AMReX codes.

- [x] test on CI for CPU
- [x] test locally on GPU (use names from #545 for uniform test name?)

### DIFF
--- a/tests/test_podvector.py
+++ b/tests/test_podvector.py
@@ -90,6 +90,50 @@ def test_from_numpy_normalizes_input():
     np.testing.assert_array_equal(result2, np.array([4.0, 5.0, 6.0], result2.dtype))
 
 
+def test_to_device_empty():
+    podv = amr.PODVector_int_std()
+    device = podv.to_device()
+    assert isinstance(device, amr.DeviceVector_int)
+    assert device.size() == 0
+    assert device.empty()
+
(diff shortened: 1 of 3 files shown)


tests/test_podvector.py
```diff
@@ -41,3 +43,86 @@ def test_array_interface():
...
+
+
+def test_from_xp_default():
+    import numpy as np
+
+    arr = np.array([5.0, 6.0, 7.0])
+    podv = amr.DeviceVector_real.from_xp(arr)
+    assert podv.size() == 3
+    result = podv.to_numpy(copy=True)
+    np.testing.assert_array_equal(result, arr)
```
EZoni: The two tests `test_from_xp` and `test_from_xp_default` seem identical, except for the numerical values of the array elements.

ax3l: Ah, because I replace-all updated it from earlier (underlying) types. Will consolidate.

ax3l: fixed.

ax3l:
Tested locally to run well on GPU :)

PR #561, opened 2026-04-20
RTD: Update Build Env

ax3l:
Update to Python 3.13 & latest Ubuntu LTS

Update stubgen with [this enum fix](https://github.com/pybind/pybind11-stubgen/pull/303).

[Preview](https://pyamrex--561.org.readthedocs.build/en/561/usage/api.html).


ax3l:
pushed directly to `development`

PR #565, opened 2026-04-24
Relax pybind11 version requirement for package managers

EZoni:
## Overview

Fix the issue raised in https://github.com/BLAST-ImpactX/impactx/pull/1419#discussion_r3128341519:

> @EZoni I overlooked something here: You do not want to update the version in `find_package` higher than needed, because that way we cannot build with version `3.0.0`, `3.0.1`, `3.0.2`, and `3.0.3` in package managers anymore - even if those versions are totally fine. (e.g., conda-forge is still to ship `3.0.4`, stands at `3.0.3`. Spack has very good version control, too). So in other words: keep this here on `3.0.0` and below just update the version we download to the latest.

## Related PRs

- https://github.com/BLAST-ImpactX/impactx/pull/1432
- https://github.com/BLAST-WarpX/warpx/pull/6805

## Notes

Please select "Hide whitespace" to review the changes or use `git diff -w`.

### DIFF
--- a/update_dependencies.py
+++ b/update_dependencies.py
@@ -63,17 +63,22 @@ def update(args):
     # loop over repositories and update dependencies data
     for repo_name, repo_subdict in repo_dict.items():
         print(f"\nUpdating {repo_labels[repo_name]}...")
+
         # set keys to access dependencies data
         commit_key = f"commit_{repo_name}"
         version_key = f"version_{repo_name}"
+
         # get new commit information
         commit_response = requests.get(repo_subdict["commit"])
         commit_dict = commit_response.json()
+
         # set new commit
         repo_commit_sha = commit_dict["sha"]
+
         # get new version tag information
         tags_response = requests.get(repo_subdict["tags"])
         tags_list = tags_response.json()
+
         # filter out old-format tags for specific repositories
         tags_list_filtered = copy.deepcopy(tags_list)
         if repo_name == "amrex":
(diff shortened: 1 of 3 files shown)


update_dependencies.py
```diff
@@ -82,33 +87,46 @@ def update(args):
...
+            # use version tag instead of commit sha:
+            # - for a release update
+            # - for pybind11 (always)
+            # - if the commit has not changed since the last version tag
+            use_version_tag = (
+                args.release
+                or (repo_name == "pybind11")
+                or (repo_commit_sha == tags_list_filtered[0]["commit"]["sha"])
+            )
+            new_commit_sha = repo_version_tag if use_version_tag else repo_commit_sha
```
EZoni, line 110: This additional change is porting the fix described in https://github.com/BLAST-WarpX/warpx/pull/6439#issuecomment-3628121530, which in turn referenced https://github.com/BLAST-WarpX/warpx/pull/6432#discussion_r2589994727:

> What would be a good logic that I did manually for the weekly commits: If in fact the latest commit is described by the latest tag, I kept using the tag in commit_* even for weekly updates. That avoids flip-flopping this line for dependencies that have releases less often. Also makes it very obvious when a dependency of ours needs a new tag during development monthly cycles.

The pyAMReX autoupdate script now matches the WarpX one more closely.

ax3l (approved):

PR #566, opened 2026-05-01
Release: 26.05

EZoni:
Depends on #567.


.github/workflows/dependencies/dependencies_gcc.sh
```diff
@@ -7,10 +7,13 @@
 
 set -eu -o pipefail
 
+# Parse GCC version from the command line
+gcc_version=${1:?Usage: $0 <gcc_version>}}
```
EZoni: This produces the following error,
```
$ ./dependencies_gcc.sh  # wrong command line call missing gcc version
./dependencies_gcc.sh: line 11: 1: Usage: ./dependencies_gcc.sh <gcc_version>
```
telling there was a problem with the command line argument 1.

.github/workflows/ubuntu.yml
```diff
@@ -287,9 +287,9 @@ jobs:
         export LD_LIBRARY_PATH=/usr/local/nvidia/lib:/usr/local/nvidia/lib64:/usr/local/cuda/lib64:${LD_LIBRARY_PATH}
         which nvcc || echo "nvcc not in PATH!"
 
-        export CC=$(which gcc-10)
-        export CXX=$(which g++-10)
-        export CUDAHOSTCXX=$(which g++-10)
+        export CC=$(which gcc-11)
+        export CXX=$(which g++-11)
+        export CUDAHOSTCXX=$(which g++-11)
```
EZoni, line 292: @ax3l @WeiqunZhang 

The CI workflows errors with
```
error: #error -- unsupported GNU version! gcc versions later than 12 are
  not supported! The nvcc flag '-allow-unsupported-compiler' can be used to
  override this version check; however, using an unsupported host compiler
  may cause compilation failure or incorrect run time execution.  Use at your
  own risk.
```
However, here I am using gcc 11, not "later than 12", so I'm a bit lost.

EZoni: Okay, there's some `gcc-10` / `g++-10` hard-coded somewhere else...

EZoni: Fixed, among other things, in [0763cd3](https://github.com/AMReX-Codes/pyamrex/pull/566/commits/0763cd3c0a245c7360cd76a5799022ad6645f3df).

.github/workflows/ubuntu.yml
```diff
@@ -249,8 +249,8 @@ jobs:
         run: |
           python3 -m pytest tests/
 
-  nvcc11:
-    name: CUDA@12.2 GNU@10.5
+  nvcc:
```
EZoni, line 252: I suggest naming the job `nvcc` instead of `nvcc11`, `nvcc12`, etc., for easier maintenance. The name displayed is anyways overwritten by the `name` field below.

.github/workflows/ubuntu.yml
```diff
@@ -267,7 +267,7 @@ jobs:
     - uses: actions/checkout@v6
     - name: Dependencies
       run: |
-        .github/workflows/dependencies/dependencies_nvcc12.sh
+        .github/workflows/dependencies/dependencies_nvcc.sh 12.2
```
EZoni, line 270: I suggest avoiding hard-coding the version number in the script's name for easier maintenance.

.github/workflows/dependencies/dependencies_nvcc.sh
```diff
@@ -11,13 +11,17 @@
 
 set -eu -o pipefail
 
+# Parse NVCC version from the command line
+nvcc_version_dotted=${1:?Usage: $0 <nvcc_version>}
+nvcc_version_dashed=${nvcc_version_dotted/./-}  # replace first occurence of "." with "-"
```
EZoni, line 16: Same command line parsing logic as in the corresponding dependency script in WarpX, except for the usage error message which is new in this PR.

.github/workflows/ubuntu.yml
```diff
@@ -57,14 +57,14 @@ jobs:
         mpiexec -np 1 python3 -m pytest tests/
 
   gcc14:
-    name: GNU@14.5 Debug
+    name: GNU@14 Debug
```
EZoni, line 60: Renamed this `GNU@14` because we install gcc/g++ 14 - not sure why it was named `GNU@14.5`.

EZoni:
The workflows
- Ubuntu / CUDA@12.2 GNU@10.5
- Ubuntu / GNU@10.5 Release
- Ubuntu / GNU@14.5 Debug

have been renamed and should be removed from the list of required workflows in the settings of the repository. The new workflows corresponding to these should be added to the list as well.

.github/workflows/ubuntu.yml
ax3l: good idea!

.github/workflows/ubuntu.yml
ax3l: good idea!

.github/workflows/dependencies/dependencies_nvcc.sh
ax3l: LGTM

ax3l:
Thank you, done!

ax3l (approved):
Thank you! LGTM! :)

PR #570, opened 2026-05-20
Keep MF alive during iteration (More Interfaces)

### LINKED ISSUE #569: MF iteration issue on WSL
On a Windows Subsystem for Linux build I ran into the following issue. Directly calling 
```py
for mfi in MF:
    ...
```
where MF is a `<amrex.MultiFab with '1' components>` object results in a segfault. The Backtrace is rather cryptic but has entries of the type `PyObject_GenericGetDict at ??:?`.

If instead I do:
```py
        import amrex.space3d as amr
        mfi = amr.MFIter(Bx)
        dir(mfi)
        while mfi.is_valid:
            try: 
                ...
                next(mfi)
            except StopIteration:
                break        
```
things work fine. It turned out the critical thing was triggering some more complete initialization via the `dir(mfi)` line, pointing to some or other incomplete initialization in the current `development`. 

As an overall fix for my WarpX sims using pyamrex, I now do the following at the start of my script:
```py
def apply_wsl2_patches():
    """
    Apply patches for WSL2 pybind11 initialization quirks.
    See: https://github.com/AMReX-Codes/amrex/issues/XXXX
    """
    import amrex.space3d as amr

    def _multifab_iter(self):
        mfi = amr.MFIter(self)
        dir(mfi)
        while mfi.is_valid:
            try:
                yield mfi
                next(mfi)
            except StopIteration:
                break
    amr.MultiFab.__iter__ = _multifab_iter

apply_wsl2_patches()
```

but it would be good to understand and fix the incomplete initialization that causes the issue in the first place. Also, it is not clear why this happens on WSL and not on "proper" Unix builds (Ubuntu or Mac).

roelof-groenewald:
Fixes #569.

Claude thinks the problem encountered on WSL is that the garbage collector is destroying needed references before the MF iteration is complete. The changes here were found to avoid the issue.

### DIFF
--- a/src/Base/Array4.H
+++ b/src/Base/Array4.H
@@ -139,9 +139,9 @@ namespace pyAMReX
             .def_property_readonly("num_comp", &Array4<T>::nComp)
 
             .def(py::init< >())
-            .def(py::init< Array4<T> const & >())
-            .def(py::init< Array4<T> const &, int >())
-            .def(py::init< Array4<T> const &, int, int >())
+            .def(py::init< Array4<T> const & >(), py::keep_alive<1, 2>())
+            .def(py::init< Array4<T> const &, int >(), py::keep_alive<1, 2>())
+            .def(py::init< Array4<T> const &, int, int >(), py::keep_alive<1, 2>())
             //.def(py::init< T*, Dim3 const &, Dim3 const &, int >())
 
             /* init from a numpy or other buffer protocol array: non-owning view
@@ -189,7 +189,7 @@ namespace pyAMReX
                 // todo: we could check and store here if the array buffer we got is read-only
 
                 return a4;
-            }))
+            }), py::keep_alive<1, 2>())
 
             /* init from __cuda_array_interface__: non-owning view
              * TODO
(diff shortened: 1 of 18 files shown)


ax3l:
Thanks for this, I will take a look!

We do have keep-alive instructions in the pybind11 bindings, but maybe one is lacking.

roelof-groenewald:
The problem might be something specific to WSL, so it's not clear to me whether this really fixes a current bug or if the "fix" is more just a hack to get WSL to behave.

ax3l:
The work-around here is pretty solid, but there is a less verbose, more fundamental way to do it directly in the `pybind11` interfaces. I will push to this PR and can you then potentially test it again? :)

ax3l:
@roelof-groenewald I added a cleaner fix for the interfaces that lacked it. But I could not reproduce the original issue at any point from #569. Is it possible the snippets there were over-reduced from another interface?

Can you try this branch again please? :)

ax3l:
I can push a few more lifetime fixes in other APIs here, maybe they cover your case.

ax3l (approved):

roelof-groenewald:
@ax3l thanks for adding the extra lifetime fixes!

roelof-groenewald:
> I added a cleaner fix for the interfaces that lacked it. But I could not reproduce the original issue at any point from #569. Is it possible the snippets there were over-reduced from another interface?

Were you trying it on a Windows system?

ax3l:
Nope. Can you please check on the system where you saw the issues?

PR #571, opened 2026-06-01
Update bindings for latest AMReX particle/vector APIs

WeiqunZhang:
Adapt pybind bindings to AMReX C++20 interface changes:
- update IntVectND::dim3 and RealVect::crossProduct
- update ParticleContainer::add_particles for requires-based addParticles
- refresh ParticleContainer comments for addParticles and WritePlotFile overloads
- update ParticleTile setSuperParticle/getSuperParticle

### DIFF
--- a/src/Base/IntVect.cpp
+++ b/src/Base/IntVect.cpp
@@ -68,7 +68,7 @@ namespace
 
         if constexpr (dim >= 1 && dim <=3) {
             py_iv.def("dim3",
-               [](const iv_type& iv) { return iv.dim3(); });
+               py::overload_cast<>(&iv_type::dim3, py::const_));
         }
 
         py_iv
--- a/src/Base/IntVect.cpp
+++ b/src/Base/IntVect.cpp
@@ -68,7 +68,7 @@ namespace
 
         if constexpr (dim >= 1 && dim <=3) {
             py_iv.def("dim3",
-               py::overload_cast<>(&iv_type::template dim3<dim>, py::const_));
+               [](const iv_type& iv) { return iv.dim3(); });
         }
 
         py_iv
--- a/src/Base/IntVect.cpp
+++ b/src/Base/IntVect.cpp
@@ -68,7 +68,7 @@ namespace
 
         if constexpr (dim >= 1 && dim <=3) {
             py_iv.def("dim3",
-               py::overload_cast<>(&iv_type::dim3, py::const_));
+               [](const iv_type& iv) { return iv.dim3(); });
         }
 
         py_iv
--- a/src/Base/RealVect.cpp
+++ b/src/Base/RealVect.cpp
@@ -93,11 +93,7 @@ void init_RealVect(py::module &m) {
           .def(py::self * py::self)
           .def("dotProduct", &RealVect::dotProduct, "Return dot product of this vector with another")
 #if (AMREX_SPACEDIM == 3)
-          .def("crossProduct",
-               [](const RealVect& lhs, const RealVect& rhs) {
-                   return lhs.crossProduct(rhs);
-               },
-               "Return cross product of this vector with another")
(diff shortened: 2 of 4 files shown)


ax3l:
Thank you, @WeiqunZhang! Shall we merge this for `26.06` of AMReX already or just afterwards? :)

WeiqunZhang:
We shall merge this after the 26.06 release.

WeiqunZhang:
Let's also wait for https://github.com/AMReX-Codes/amrex/pull/5468. A few more changes are needed.

ax3l:
Read for rebase! :)

pyAMReX release is out, too.

src/Base/IntVect.cpp
```diff
@@ -68,7 +68,7 @@ namespace
 
         if constexpr (dim >= 1 && dim <=3) {
             py_iv.def("dim3",
-               py::overload_cast<>(&iv_type::template dim3<dim>, py::const_));
+               [](const iv_type& iv) { return iv.dim3(); });
```
ax3l, line 71: Is writing them as verbose lambdas the only way? Or can we still use `&method` and `overload_cast` w/ the update templates?

I prefer not writing "trivial" lambdas, when possible, to keep the code less verbose.

(Applies to all changes in this PR.)

WeiqunZhang: We can use the current style for dim3, crossProduct, and addParticles. But it appears difficult for getSuperParticle and setSuperParticle, because the overloads have the same signatures other than the "requires" clause, which is not part of the function signature.

ax3l: Makes sense, yes let's try to use the short form where we can.

WeiqunZhang: /home/runner/work/pyamrex/pyamrex/src/Base/IntVect.cpp(71): error: no instance of overloaded function "pybind11::detail::overload_cast_impl<Args...>::operator() [with Args=<>]" matches the argument list
            argument types are: (<unknown-type>, const std::true_type)
            object type is: const pybind11::detail::overload_cast_impl<>
                 py::overload_cast<>(&iv_type::dim3, py::const_));
                 ^
          detected during instantiation of "void <unnamed>::init_IntVectND<dim>(pybind11::module &) [with dim=1]" at line 181

/home/runner/work/pyamrex/pyamrex/src/Base/IntVect.cpp(71): error: no instance of overloaded function "pybind11::detail::overload_cast_impl<Args...>::operator() [with Args=<>]" matches the argument list
            argument types are: (<unknown-type>, const std::true_type)
            object type is: const pybind11::detail::overload_cast_impl<>
                 py::overload_cast<>(&iv_type::dim3, py::const_));
                 ^
          detected during instantiation of "void <unnamed>::init_IntVectND<dim>(pybind11::module &) [with dim=2]" at line 182

/home/runner/work/pyamrex/pyamrex/src/Base/IntVect.cpp(71): error: no instance of overloaded function "pybind11::detail::overload_cast_impl<Args...>::operator() [with Args=<>]" matches the argument list
            argument types are: (<unknown-type>, const std::true_type)
            object type is: const pybind11::detail::overload_cast_impl<>
                 py::overload_cast<>(&iv_type::dim3, py::const_));
                 ^
          detected during instantiation of "void <unnamed>::init_IntVectND<dim>(pybind11::module &) [with dim=3]" at line 183

It worked on my machines using GCC 13. But Clang 18 and GCC 12 don't like it.

WeiqunZhang: I think we have to use a lambda function here.

WeiqunZhang: Looks like we have issues wit RealVect::crossProduct too. Maybe DeepSeek did know what it was doing, and we have to go back to lambda.

ax3l: Okay then. Thanks for trying!

ax3l (approved):

PR #573, opened 2026-06-02
CI: Add Windows on Ubuntu WLS

ax3l:
Add a Windows WSL runner that runs Ubuntu.


ax3l:
Ah, will only see after merge...

PR #581, opened 2026-06-05
Particle Reading (Runtime SoA)

ax3l:
Add methods to read into a general polymorpic, all-runtime PC from plotfiles (output or checkpoint).
Related to https://github.com/AMReX-Codes/pyamrex/discussions/488

- [x] review draft
- [x] depends on https://github.com/AMReX-Codes/amrex/pull/5476
- [x] add a user-facing workflow page: ["How to read back plotfiles"](https://pyamrex--581.org.readthedocs.build/en/581/usage/workflows.html)

### DIFF
--- a/src/amrex/extensions/ParticleContainer.py
+++ b/src/amrex/extensions/ParticleContainer.py
@@ -171,6 +171,82 @@ def pc_to_df(self, local=True, comm=None, root_rank=0):
     return df
 
 
+def read_particles(
+    amr, plotfile, particle_dir="particles", communicate=True, container=None
+):
+    """Read AMReX particle data from a plotfile or checkpoint.
+
+    The on-disk layout (number and names of the real/int components, the
+    precision and whether the file is a checkpoint) is discovered via
+    :py:class:`amrex.ParticleHeader`, which uses the same AMReX C++ header reader
+    as ``ParticleContainer::Restart``. A polymorphic, pure Struct-of-Arrays
+    ParticleContainer is then configured with matching runtime components and
+    restarted from the file - so no prior knowledge of the container's
+    compile-time layout is required.
+
+    Parameters
+    ----------
+    amr : module
+        The dimension-specific pyAMReX module (e.g. ``amrex.space3d``).
+    plotfile : str
+        Path to the plotfile / checkpoint directory.
+    particle_dir : str, optional
+        Name of the particle sub-directory inside ``plotfile``
+        (default: ``"particles"``).
+    communicate : bool, optional
+        Whether the added runtime components participate in redistribution
+        (default: ``True``).
+    container : ParticleContainer, optional
+        An existing, geometry-defined container to restart into. If ``None``
+        (default), the geometry is recovered from the plotfile's
+        :py:class:`amrex.PlotFileData` and a fresh polymorphic pure-SoA
+        container is created.
+
+    Returns
+    -------
+    ParticleContainer
+        The populated particle container. Iterate the data via, e.g.,
+        ``for pti in pc.iterator(level=0): soa = pti.soa()``.
+    """
+    header = amr.ParticleHeader.read(plotfile, particle_dir)
+
+    pc = container
+    if pc is None:
+        # recover the AMR geometry from the plotfile metadata
+        plt = amr.PlotFileData(plotfile)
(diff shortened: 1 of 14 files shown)


BenWibking (commented):
Looks reasonable to me. I am building this PR now to test with some Quokka plotfiles.

src/amrex/extensions/ParticleContainer.py
```diff
@@ -171,6 +171,82 @@ def pc_to_df(self, local=True, comm=None, root_rank=0):
...
+    ParticleContainer
+        The populated particle container. Iterate the data via, e.g.,
+        ``for pti in pc.iterator(level=0): soa = pti.soa()``.
+    """
+    header = amr.ParticleHeader.read(plotfile, particle_dir)
+
+    pc = container
+    if pc is None:
+        # recover the AMR geometry from the plotfile metadata
+        plt = amr.PlotFileData(plotfile)
```
BenWibking, line 259: Are particle outputs guaranteed to have a plotfile `Header`? It should probably throw an exception if this doesn't exist.

BenWibking:
Is this intended to work for multi-level plotfiles?

src/amrex/extensions/ParticleContainer.py
ax3l: Good questions for @atmyers :)

ax3l: Claude tells me:

>  - `<plotfile>/Header` (line 216, amr.PlotFileData(plotfile)): not guaranteed. It's only written by the mesh plotfile writers (WriteSingleLevelPlotfile etc.). A particle-only output — e.g.
  pc.write_plotfile("dir", "particles", ...) into a fresh directory, or a bare pc.checkpoint — has no top-level Header. When it's missing, `PlotFileDataImpl`'s constructor goes through `ReadAndBcastFile`
  → `FileOpenFailed` → `amrex::Abort`, which kills the Python interpreter — only our test suite sets `amrex.throw_exception=1` (tests/conftest.py:71); normal user sessions don't.
>   - `<plotfile>/<particle_dir>/Header` (line 211, `ParticleHeader.read`): always written by the particle writers, but a typo'd path dies the same way — `ParticleHeader::read` also uses `ReadAndBcastFile` with
  no existence check.
>
>  There's a subtler catch, too: for an application checkpoint directory the top-level Header exists but is in checkpoint format, which `PlotFileData` cannot parse — so existence alone isn't sufficient
  for the container=None path.
>
>  Recommended fix: `os.path.isfile` checks raising `FileNotFoundError` with actionable messages — always for the particle Header, and in the container is None branch for the top-level Header, telling the
  user to pass `container=` (a geometry-defined container) when the geometry can't be recovered from the file. Document in the docstring that geometry auto-recovery requires a plotfile with mesh data;
  checkpoints need `container=`.

atmyers: Particle outputs must have a plotfile `Header`. We support pure particle outputs without mesh data by writing a dummy MultiFab so that this `Header` gets included.

ax3l: yay, thanks!

atmyers: Claude is correct that it's possible write particle output without a plotfile, but most/all of the visualization tools won't understand them without the `Header`, so this shouldn't be done in practice.

ax3l:
Let's add a test for that. I think reading should work, but we might add them simply on level 0 for now an let users redistribute as needed.

ax3l:
Change:
  - Particles from all levels are read and when no container is passed, the auto-created container is single-level and all particles are placed on level 0 (positions preserved). There's a two-level roundtrip test now.
  - Added explicit checks: missing particle or plotfile Header now raises `FileNotFoundError`, and checkpoint-format top-level Headers raise `ValueError` pointing to the container argument, since `PlotFileData` can't parse those.

ax3l:
@atmyers @BenWibking ready for review / merge :)

ax3l:
@roelof-groenewald @WeiqunZhang Andrew and Ben are not available for a few week(s), can you potentially approve? :)

roelof-groenewald (approved):
Cool new feature!

PR #595, opened 2026-07-18
Migrate to nanobind, remove pybind11

BenWibking:
This is a vibe-coded experiment to see if it is feasible to replace pybind11 with nanobind. Apparently, it is!

[Benefits](https://github.com/wjakob/nanobind#nanobind-tiny-and-efficient-cpython-bindings):
* "~4× faster compile time"
* "~5× smaller binaries"
* " ~10× lower runtime overheads compared to pybind11"
* **Hidden bonus feature:** DLPack batteries included via [`nb::ndarray<...>`](https://nanobind.readthedocs.io/en/latest/ndarray.html#array-libraries)

Generated using ~1 million tokens from GPT-5.6-Sol.


ax3l:
Cool! @BenWibking van you try to use nanobind's stub generator to replace our pybind11-stubgen? We need stubs for docs and ipython completion.

CMakeLists.txt
```diff
@@ -366,6 +373,24 @@ add_custom_target(pyAMReX_python_sources ALL
...
+if(pyAMReX_BUILD_STUBS)
+    add_custom_target(pyAMReX_stubs)
+    foreach(D IN LISTS AMReX_SPACEDIM)
+        nanobind_add_stub(pyAMReX_${D}d_stub
+            MODULE amrex_${D}d_pybind
+            OUTPUT
+                ${pyAMReX_SOURCE_DIR}/src/amrex/space${D}d/amrex_${D}d_pybind/__init__.pyi
+            PYTHON_PATH $<TARGET_FILE_DIR:pyAMReX_${D}d>
+            DEPENDS pyAMReX_${D}d
+        )
```
BenWibking, line 389: @ax3l nanobind stubs are generated here using nanobind's CMake interface: https://nanobind.readthedocs.io/en/latest/typing.html#cmake-interface

BenWibking:
Amazingly, all tests pass on all compilers. I'm marking it ready for review.

src/Base/FabArray.cpp
```diff
@@ -19,87 +19,87 @@
 namespace
 {
     template<typename T>
-    void make_FabArray_T(py::module &m, std::string const &name)
+    void make_FabArray_T(nb::module_ &m, std::string const &name)
```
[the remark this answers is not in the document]
BenWibking: I don't understand this comment. @ax3l do you?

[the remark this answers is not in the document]
ax3l: Huh, I think that is a false positive.

tests/downstream_nanobind/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+cmake_minimum_required(VERSION 3.24)
+project(pyAMReXDownstreamNanobind LANGUAGES C CXX)
+
+find_package(Python 3.8 COMPONENTS Interpreter Development.Module REQUIRED)
```
ax3l, line 4: Oh, that's a neat downstream integration test.

CMakeLists.txt
```diff
@@ -366,6 +373,24 @@ add_custom_target(pyAMReX_python_sources ALL
...
+        add_dependencies(pyAMReX_stubs pyAMReX_${D}d_stub)
```
ax3l, line 390: We might need to add a dependency `pyAMReX_${D}d_stub` -> `pyAMReX_${D}d` as well.

Maybe also env var helpers in case we do not want to rely on a  `pip_install` (to check).

setup.py
```diff
@@ -45,10 +45,19 @@ def run(self):
...
+            ignore=shutil.ignore_patterns(
+                "__pycache__", "*.pyc", "chk*", "diags*", "plt*"
+            ),
         )
 
 
+class BinaryDistribution(Distribution):
+    """Keep prebuilt-artifact wheels correctly tagged as platform wheels."""
+
+    def has_ext_modules(self):
```
ax3l: We probably can already merge those install fixes separately.

setup.py
```diff
@@ -273,6 +291,7 @@ def build_extension(self, ext):
     },
     # CMake: self-built as extension module
     ext_modules=cxx_modules,
+    distclass=BinaryDistribution,
```
ax3l: See above: probably separate PR already for this (fix)?

ax3l (commented):
This looks pretty great!

I added a few inline thoughts to research.

I will push a commit to undo the `.pyi` changes and might isolate the `setup.py` binary distr update once I understand it to a separate PR.

tests/downstream_nanobind/CMakeLists.txt
BenWibking: The AI came up with that after it broke on an earlier change. I think it's useful.

BenWibking:
I was confused to whether the .pyi files should be checked in. What is the policy on that? Is it documented somewhere?

ax3l:
Sorry for the confusion.

Yes, the `.pyi` files are checked in, because we need them to build Sphinx docs from.
(We have not enough CPU hours to rebuild the C++ bindings in RTD CI builders.)
Additionally, tools like IPython's Jedi auto-complete need them to give interactive users auto-completion hints.

But, they are only checked in via an auto-commit on `development`.
The reason for that is that if PRs change them, then nearly every PR will have a merge conflict with other open PRs. (I tried for a few days and realized that does not scale.)

Thus, our PR's CIs do build them, check they are valid, but they do not commit them.
Only the `development` branch does the last step, too: commit on changes.

yes, we should document this workflow / automation in the `development` or `maintenance` sections of the docs:
https://pyamrex.readthedocs.io

BenWibking:
> Sorry for the confusion.
> 
> Yes, the `.pyi` files are checked in, because we need them to build Sphinx docs from.
> 
> But, they are only checked in via an auto-commit on `development`. The reason for that is that if PRs change them, then nearly every PR will have a merge conflict with other open PRs. So that does not scale.

ah, okay :) That makes sense.

ax3l:
Before we merge this in (maybe early in a release cycle), I would do a whole BLAST build test to see if mixing nanobind and pybind11 works well for the mid-term, then migrate the rest of BLAST too. (I expect no issues.)

BenWibking:
Ok, I can extract the build fixes to a separate PR. I have some uncommitted fixes to make things easier when using `uv` to build/install.

BenWibking:
Build fixes moved to https://github.com/AMReX-Codes/pyamrex/pull/599.

setup.py
BenWibking: Moved to https://github.com/AMReX-Codes/pyamrex/pull/599.

setup.py
BenWibking: Moved to https://github.com/AMReX-Codes/pyamrex/pull/599.

PR #599, opened 2026-07-29
Add wheel build isolation option and fix platform tagging

BenWibking:
## Summary

Separates wheel-install reliability fixes from the nanobind migration.

- add an option to enable pip build isolation for the CMake wheel target
- mark prebuilt artifact wheels as platform wheels
- exclude Python bytecode caches from staged package contents

## Why

The wheel target often needs its declared build requirements installed in an isolated environment, and a wheel containing compiled artifacts must receive a platform-specific tag.

## Validation

```console
$ uv venv --python 3.13 .venv
$ source .venv/bin/activate
$ AMREX_SPACEDIM=3 \
   CMAKE_BUILD_PARALLEL_LEVEL=8 \
   uv pip install -e .
$ python -c 'import amrex.space3d as amr; print(amr.Config.spacedim)'
3
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -311,7 +311,7 @@ add_custom_target(${pyAMReX_CUSTOM_TARGET_PREFIX}pip_wheel
     ${CMAKE_COMMAND} -E rm -f -r amrex-whl
     COMMAND
         ${CMAKE_COMMAND} -E env PYAMREX_LIBDIR=${CMAKE_PYTHON_OUTPUT_DIRECTORY}/amrex/
-            ${Python_EXECUTABLE} -m pip ${PY_PIP_OPTIONS} wheel --no-build-isolation --no-deps --wheel-dir=amrex-whl "${pyAMReX_SOURCE_DIR}"
+            ${Python_EXECUTABLE} -m pip ${PY_PIP_OPTIONS} wheel --no-deps --wheel-dir=amrex-whl "${pyAMReX_SOURCE_DIR}"
     COMMAND_EXPAND_LISTS VERBATIM
     WORKING_DIRECTORY
         ${pyAMReX_BINARY_DIR}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -296,6 +296,15 @@ set(PY_PIP_OPTIONS "-v" CACHE STRING
     "Additional parameters to pass to `pip` as ; separated list")
 set(PY_PIP_INSTALL_OPTIONS "" CACHE STRING
     "Additional parameters to pass to `pip install` as ; separated list")
+option(pip-build-isolation
+    "Build the Python wheel in an isolated build environment"
+    OFF
+)
+if(pip-build-isolation)
+    set(_pyAMReX_PIP_BUILD_ISOLATION_OPTION "")
+else()
+    set(_pyAMReX_PIP_BUILD_ISOLATION_OPTION "--no-build-isolation")
+endif()
 
 # add a prefix to custom targets so we do not collide if used as a subproject
 if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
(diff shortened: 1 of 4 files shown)


BenWibking:
@ax3l What is the motivation for `--no-build-isolation`?

ax3l:
Build isolation:
I have to remind myself about all the details, but at least we need to depend on system packages on HPC...

BenWibking:
> Build isolation: I have to remind myself about all the details, but at least we need to depend on system packages on HPC...

Can we avoid that? We only actually need to link against the system-provided MPI.

If it can be self-contained except for MPI, it would be much easier to make it installable via package managers like `pixi` (and reproducible). Also, if it can be self-contained, we won't need a separate module recipe for each HPC machine. That makes it simpler.

ax3l:
I think we probably can avoid it, but I have to dig through the history to find the quirks that might explore if we take it off :D

Let's do some testing after the `26.08` release, maybe some of the tooling problems of 2022 are now overcome.

> We only actually need to link against the system-provided MPI.

Well, we want to include system headers, honor system compiler flags on HPC in CXXFLAGS, link against shared libs, etc. it's kind of the opposite of portability once we reach an HPC machine :sweat: 
With the latest ABI compatible MPI 5 this gets a bit better.

BenWibking:
> I think we probably can avoid it, but I have to dig through the history to find the quirks that might explore if we take it off :D
> 
> Let's do some testing after the `26.08` release, maybe some of the tooling problems of 2022 are now overcome.
> 
> > We only actually need to link against the system-provided MPI.
> 
> Well, we want to include system headers, honor system compiler flags on HPC in CXXFLAGS, link against shared libs, etc. it's kind of the opposite of portability once we reach an HPC machine 😓 With the latest ABI compatible MPI 5 this gets a bit better.

Ok, I guess that's true. I have successfully run on Frontier with a container that bind-mounts Cray MPI and libcxi, but it is matched to a compatible OpenSUSE release. (I should test if it works with a different distro in the container, but haven't had time...)

This is what I have bind mounted on Frontier:
```
## import shared libraries from host filesystem
export APPTAINER_BINDPATH=/usr/share/libdrm,/usr/share/libdrm/amdgpu.ids:/opt/amdgpu/share/libdrm/amdgpu.ids,/var/spool/slurmd,/opt/cray,${PWD}
export APPTAINERENV_LD_LIBRARY_PATH="$CRAY_MPICH_ROOTDIR/gtl/lib:/opt/rocm/lib:/opt/rocm/lib64:$CRAY_LD_LIBRARY_PATH:$LD_LIBRARY_PATH:/opt/cray/pe/lib64"
export APPTAINER_CONTAINLIBS="/usr/lib64/libcxi.so.1,/usr/lib64/libjson-c.so.5,/lib64/libtinfo.so.6,/usr/lib64/libnl-3.so.200,/usr/lib64/libgfortran.so.5,/usr/lib64/libjansson.so.4,/usr/lib64/libatomic.so.1"
export APPTAINERENV_LD_PRELOAD=$CRAY_MPICH_ROOTDIR/gtl/lib/libmpi_gtl_hsa.so.0
```

Hopefully, only these libs are necessary to link against. From my point of view, it would be ideal if we could essentially containerize it with Pixi, except for these libs.

ax3l:
Ah, let's not bring containers as a build _requirement_ into the game :D
I love them, but even the three big DOE sites each have a different bespoke solutions, and they are not always performance neutral. We must be able to also compile bare-metal.

For cleanliness & package maintainers, I think we can add an option to disable build isolation (but keep building un-isolated by default for now.)

Big pixi fan, so adding on that end I am def also very excited about.

BenWibking:
> Ah, let's not bring containers as a requirement into the game :D I love them, but even the three big DOE sites each have a different solution. We must be able to also compile bare-metal.

I didn't mean to suggest requiring containers. I just point to it as a proof-of-principle that you can avoid linking to anything except for MPI and libcxi and get a hermetic build.

Long term, I view being able to do a hermetic build as a positive thing, since then we can distribute packaged binaries that are `pixi add`-able, and once MPI-5 ABI is available, avoid having to compile on HPC machines at all.

CMakeLists.txt
```diff
@@ -311,7 +311,7 @@ add_custom_target(${pyAMReX_CUSTOM_TARGET_PREFIX}pip_wheel
     ${CMAKE_COMMAND} -E rm -f -r amrex-whl
     COMMAND
         ${CMAKE_COMMAND} -E env PYAMREX_LIBDIR=${CMAKE_PYTHON_OUTPUT_DIRECTORY}/amrex/
-            ${Python_EXECUTABLE} -m pip ${PY_PIP_OPTIONS} wheel --no-build-isolation --no-deps --wheel-dir=amrex-whl "${pyAMReX_SOURCE_DIR}"
```
ax3l, line 314: Control via a build option `pip-build-isolation` with default `OFF` for now?

BenWibking: That makes sense to me.

docs/source/install/cmake.rst
```diff
@@ -91,6 +91,7 @@ CMake Option                    Default & Values                             Des
 ``AMReX_SPACEDIM``              ``3``                                        Dimension(s) of AMReX as a ``;``-separated list
 ``AMReX_BUILD_SHARED_LIBS``     ON/**OFF**                                   Build AMReX library as shared (required for app extensions)
 ``pyAMReX_INSTALL``             **ON**/OFF                                   Enable install targets for pyAMReX
+``pip-build-isolation``         ON/**OFF**                                   Build the Python wheel in an isolated build environment
```
BenWibking, line 94: @ax3l Let me know what naming convention you want to use here.

BenWibking:
@ax3l ping

PR #600, opened 2026-07-30
Config: Refactor into Dedicated Module

ax3l:
Port the improved Config treatment from ImpactX
(BLAST-ImpactX/impactx#1580) to pyAMReX:

- move the `Config` bindings from `AMReX.cpp` into a new `Config.cpp`
- centralize all build-time options in a single `std::variant`-based map, from which the read-only static properties are generated
- add a custom metaclass `__repr__` that prints the build configuration as an aligned table
- add `Config.to_dict()` returning the build configuration as a dict
- keep the runtime-mutable `verbose` option as a separate read/write static property

### DIFF
--- a/src/Base/AMReX.cpp
+++ b/src/Base/AMReX.cpp
@@ -3,14 +3,9 @@
 #include <AMReX.H>
 #include <AMReX_Vector.H>
 #include <AMReX_ParmParse.H>
-#include <AMReX_SIMD.H>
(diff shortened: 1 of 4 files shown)


roelof-groenewald (approved):
Just clean-up and reorganization. I'm assuming you don't really care about the bot complaining about lack of inline comments.

PR #604, opened 2026-08-04
CI: Actions Cache Hygiene

ax3l:
## Problem

The repository sits at the **10 GB GitHub Actions cache quota** (9.57 GB when measured), so GitHub LRU-evicts entries. The `stubs` job loses its cache every time: `ci.yml` serializes `stubs → ubuntu → everything else`, so its ~58 MB entry is both the smallest and the oldest-accessed by the time the rest of CI has filled the quota.

Measured on PR #603 versus the last `development` run:

| | `Set Up Cache` | `Build & Install` | ccache hits |
|---|---|---|---|
| cache evicted (PR #603) | 1 s, no restore | **21m54s** | — |
| cache present (development) | 2 s, 58 MB | **46 s** | 596/599 (99.5%) |

`gh cache list --key "ccache-👑 CI-stubs"` returned nothing at all — every generation had been evicted, even though each run kept saving one.

## Changes

Reclaim quota and align the cache handling with WarpX, which does not have this problem:

- **Only run on pushes to `development`.** Developers work from forks, whose pushes never trigger this repository; the only branches pushed here directly are mainline and the pre-commit.ci / dependabot update branches, which open PRs anyway. Running on those pushes as well stored a second, redundant cache set under `refs/heads/<branch>` next to the `refs/pull/<n>/merge` one — 2.02 GB when measured. A branch filter also excludes tag pushes, which cached under the never-restorable `refs/heads/refs/tags/<tag>` (382 MB when measured).

- **Clean up the CodeQL caches like every other runner.** The scheduled `🔎 CodeQL` runs are standalone, so their caches are keyed `ccache-🔎 CodeQL-...` rather than `ccache-👑 CI-...` and were pruned by nothing — `CleanUpCache` only triggered on `👑 CI`. Additionally prune the TRAP database caches that `github/codeql-action` stores itself (~1 GB per generation, 1.93 GB when measured); those carry neither the `ccache-` prefix nor a `-git-` separator, so they need their own pass.

- **Update ccache 4.8 → 4.13.6 and `actions/cache` v4 → v6**, matching WarpX.

## Effect

Together with a one-off manual purge of the duplicates described above, cache usage went from 9.57 GB to 6.84 GB, and PR #603's next run stored a `ccache-👑 CI-stubs-*` entry again — the first one to exist in the repository during this investigation.

## Notes

- `CCACHE_MAXSIZE` values are deliberately left alone. WarpX caps at 600 M, but the `gcc10`/`stubs` jobs here build `AMReX_SPACEDIM="1;2;3"`; shrinking their budget would trade eviction at the GitHub level for thrashing inside ccache.
- The `2:distributionDirectory_Ubuntu_24.4.4` cache (371 MB, from the oneAPI setup) is still unmanaged. Left for a follow-up.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,17 @@
 name: 👑 CI
 
-on: [push, pull_request]
+on:
+  # Developers work from forks, whose pushes never trigger us here; the only
+  # branches pushed directly to this repo are mainline and the pre-commit.ci /
+  # dependabot update branches, which open PRs and are covered by the
+  # pull_request trigger below. Running on those pushes as well would only
+  # duplicate the CI and store a second, redundant set of Actions caches under
+  # refs/heads/<branch> next to the refs/pull/<n>/merge ones.
+  # A branch filter also excludes tag pushes, which would otherwise cache under
+  # the never-restorable refs/heads/refs/tags/<tag>.
+  push:
+    branches: [development]
+  pull_request:
 
 concurrency:
   group: ${{ github.ref }}-${{ github.head_ref }}-ci
--- a/.github/workflows/cleanup-cache.yml
+++ b/.github/workflows/cleanup-cache.yml
@@ -2,7 +2,9 @@ name: CleanUpCache
 
 on:
   workflow_run:
-    workflows: [👑 CI]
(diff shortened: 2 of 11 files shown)


ax3l:
LGTM

PR #609, opened 2026-08-12
Add `amr.xp`; factor the per-dimension module setup into a helper

ax3l:
Two commits: the feature, then the deduplication it motivated.

## 1. `amr.xp`: the array namespace matching the build

The `to_xp()` methods return NumPy, CuPy or dpnp arrays depending on how
pyAMReX was built, but there is no way to reach the *array library itself*.
CPU/GPU agnostic code regularly needs it for calls that are not methods on an
array:

```python
amr.xp.sin(src, out=dst)
amr.xp.meshgrid(*axes, indexing="ij")
```

Without it every downstream script re-implements the same dispatch. We already
resolve exactly this internally — `dlpack_helpers.xp_module_name()` returns
`"numpy"`/`"cupy"`/`"dpnp"` and the `to_xp` methods dispatch on it — so this
exposes the module that name refers to, keeping one source of truth.

### CuPy and dpnp remain optional dependencies

`amr.xp` is resolved on **first attribute access** via a PEP 562 module
`__getattr__`, never at import, so `import amrex` on a CUDA or SYCL build still
works with neither installed. After the first access the module is cached in
the module globals, bypassing `__getattr__` thereafter.

There is a regression test for this, run in a subprocess so it holds
independently of what the rest of the session imported:

```python
assert "cupy" not in sys.modules
assert "dpnp" not in sys.modules
```

If the library really is missing, the error names it rather than surfacing a
bare `ModuleNotFoundError` for a name the user never typed:

```
ImportError: amrex.xp needs 'cupy', which is an optional dependency of
pyAMReX and is not installed. Install it, or use the
to_numpy()/to_cupy()/to_dpnp() methods directly.
```

Tests `importorskip` the optional library, matching `test_podvector.py`.
`requirements.txt` and `setup.py` are untouched. The module `__getattr__` only
intercepts `xp`; anything else raises the normal `AttributeError` (tested).

## 2. Factor the per-dimension module setup into a helper

`space1d/2d/3d/__init__.py` were three near-identical copies differing only in
the pybind module they wrap and in `d_decl()`. Adding `amr.xp` would have meant
writing the same block a third time, so everything dimensionality-independent
moves to `amrex._module_api.setup_module()` and each `__init__.py` keeps only
its star-import and `d_decl()`. **About 130 lines become 25 per file.**

Module-level names cannot be installed from the `register_*_extension()`
functions the way class methods are, because `from .amrex_?d_pybind import *`
has already run by then; the helper writes them into `globals()` instead.
Injected callables get `__module__` set to the target module so that Sphinx
`autofunction` and the CI stub generator keep attributing them to
`amrex.space3d`.

### API impact

The public API is unchanged: `Print`, `d_decl`, `read_particles` and
`list_particle_species` keep their signatures, docstrings and `__module__`.
Verified by diffing every callable of `amrex.space{1,2,3}d` before and after.

The only differences are that seven `register_*_extension` names and
`_read_particles` no longer leak into the `amrex.space?d` namespace. They are
internal, documented nowhere, and WarpX imports its own `register_warpx_*` from
`pywarpx.extensions` rather than these. Expect the CI-regenerated `.pyi` stubs
to drop them.

## Testing

`ctest --test-dir build --output-on-failure` green on a
`AMReX_SPACEDIM="1;2;3" AMReX_OMP=ON` build, plus per-dimensionality checks of
`amr.xp`, `d_decl` and the re-exported helpers in all three modules. CPU only
locally — the CuPy and dpnp branches go through the shared `xp_module_name()`
path but are not run on device here.

### DIFF
--- a/src/amrex/_module_api.py
+++ b/src/amrex/_module_api.py
@@ -1,0 +1,129 @@
+"""
+This file is part of pyAMReX
+
+Copyright 2026 AMReX community
+Authors: Axel Huebl
+License: BSD-3-Clause-LBNL
+"""
+
+from .extensions.Array4 import register_Array4_extension
+from .extensions.ArrayOfStructs import register_AoS_extension
+from .extensions.MultiFab import register_MultiFab_extension
+from .extensions.ParticleContainer import (
+    list_particle_species,
+    read_particles,
+    register_ParticleContainer_extension,
+)
+from .extensions.PODVector import register_PODVector_extension
+from .extensions.SmallMatrix import register_SmallMatrix_extension
+from .extensions.StructOfArrays import register_SoA_extension
+
+
+def setup_module(ns, amr):
+    """Populate an ``amrex.space{1,2,3}d`` namespace.
+
+    Those three packages are identical apart from the compiled pybind module
+    they wrap and their ``d_decl()``, so everything else is defined once here
+    and installed into their namespace.
+
+    Class-level additions could equally be done from the ``register_*``
+    functions, because a class object is shared. Module-level names cannot:
+    ``from .amrex_?d_pybind import *`` has already run by then, so a name added
+    to the pybind module afterwards would not appear in the package namespace.
+    They have to be written into ``ns`` instead, which is what this does.
+
+    Injected callables get their ``__module__`` set to the target module, so
+    that Sphinx ``autofunction`` and the CI stub generator attribute them to
+    ``amrex.space3d`` rather than to this helper.
+
+    Parameters
+    ----------
+    ns : dict
+        The calling module's ``globals()``.
+    amr : module
+        That module's compiled bindings, e.g. ``amrex_3d_pybind``.
+    """
+    name = ns["__name__"]
+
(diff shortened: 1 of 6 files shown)


src/amrex/_module_api.py
```diff
@@ -0,0 +1,129 @@
...
+
+    read_particles_.__name__ = "read_particles"
+    read_particles_.__qualname__ = "read_particles"
+
+    def module_getattr(attr):
+        """Resolve ``xp`` lazily (PEP 562).
+
+        ``amr.xp`` is the array namespace matching this build: NumPy on CPU,
+        CuPy for CUDA/HIP, dpnp for SYCL. It is the module counterpart of the
+        ``to_xp`` methods, for code that needs to call into the array library
```
ax3l, line 87: Regression:
With this module-level `__getattr__`, _stubs_ have silently stopped catching typos on `amrex.space1d/2d/3d`, e.g., `amr.MultiFabbbb`.

Fix in #618

Issue #613, opened 2026-08-16
Release the GIL around long-running AMReX calls (`py::call_guard<py::gil_scoped_release>()`)


ax3l:
Split out of #612 (free-threaded Python). That issue is about declaring `py::mod_gil_not_used()` so free-threaded interpreters do not re-enable the GIL. This one is the complementary, *orthogonal* piece: on **GIL-enabled** builds (3.11-3.14, i.e. every wheel we ship today except `cp313t`/`cp314t`), a Python thread that calls into a long-running AMReX routine holds the GIL for the call's entire duration, so no other Python thread can run — even though the C++ code touches no Python objects.

`py::gil_scoped_release` is a no-op on free-threaded builds, so this is purely a win for the GIL builds, and it is independent of the #612 thread-safety audit.

## What it would look like

```cpp
.def("fill_boundary",
     py::overload_cast<Periodicity const &, bool>(&FAT::template FillBoundary<value_type>),
     py::call_guard<py::gil_scoped_release>(),
     ...)
```

## Candidate call sites

Expensive, callback-free entry points, e.g.:

- `FabArray::FillBoundary` — `src/Base/FabArray.cpp:230-263` (6 overloads)
- `FabArray::SumBoundary` — `src/Base/FabArray.cpp:168-181` (4 overloads)
- `FabArray::sum` — `src/Base/FabArray.cpp:162`
- `MultiFab` reductions `min`/`max`/`norm0`/`norm1`/`norm2`/`sum` — `src/Base/MultiFab.H:148-220`
- `ParticleContainer::Redistribute` — `src/Particle/ParticleContainer.H:346,350`
- `ParticleContainer::InitRandom` / `InitRandomPerBox` — `src/Particle/ParticleContainer.H:568,575`
- `write_single_level_plotfile` / `write_multi_level_plotfile` — `src/Base/PlotFileUtil.cpp:20,27`

## Why this needs its own audit

The guard is only safe where the C++ call cannot re-enter Python. Sites that **must be excluded** unless they re-acquire:

- `PyAmrCore`'s five `PYBIND11_OVERRIDE_PURE_NAME` trampolines — `src/AmrCore/AmrCore.cpp:53,64,74,84,91`. `ErrorEst` and friends call back into Python.
[...]

PR #614, opened 2026-08-17
Free-Threaded Python: Declare `py::mod_gil_not_used()`

### LINKED ISSUE #612: Free-threaded Python: declare `Py_mod_gil` (or document that the GIL is re-enabled)
We publish `cp313t`/`cp314t` wheels for ImpactX that bundle pyAMReX, but the bindings never declare free-threading support, so importing them **re-enables the GIL at runtime** and the free-threaded interpreter gives no benefit.

## Current state

`src/pyAMReX.cpp` uses the two-argument form for all three dimensionalities:

```cpp
PYBIND11_MODULE(amrex_1d_pybind, m) {
PYBIND11_MODULE(amrex_2d_pybind, m) {
PYBIND11_MODULE(amrex_3d_pybind, m) {
```

From the CPython docs ([free-threading extensions howto](https://docs.python.org/3/howto/free-threading-extensions.html)):

> Extension modules need to explicitly indicate that they support running with the GIL disabled; otherwise importing the extension will raise a warning and enable the GIL at runtime.

So `import amrex.space3d` on a free-threaded build silently turns the GIL back on process-wide (and warns under `-X warn_default_gil`). The wheels are correct and usable — they just are not free-threaded.

## What the opt-in would look like

pybind11 3.1.0 is already pinned, so the mechanism is available — pass the tag as the third argument:

```cpp
PYBIND11_MODULE(amrex_3d_pybind, m, py::mod_gil_not_used()) {
```

## Why this is not just a one-token change

pybind11 is explicit that this is a promise, not a flag ([docs](https://pybind11.readthedocs.io/en/stable/advanced/misc.html)):

> enabling your module to be used with free-threading is also your promise that your code is thread safe

That promise is not obviously true today. Worth auditing before declaring it:

- AMReX process-global state: `amrex::Initialize`/`Finalize`, the Arena / memory pools, `ParmParse`, `TinyProfiler`.
- pyAMReX hands out raw views into AMReX-owned memory (`Array4`, `MultiFab`, `PODVector`, SoA/AoS), which multiple Python threads could then mutate concurrently.
- Downstream bindings (ImpactX, WarpX) embed the same state, so whatever we decide here propagates.

Declaring the tag without that audit turns "no benefit" into "silent data races", so the sequencing matters: audit first, tag second.

## Options

1. Leave as is and document that free-threaded builds fall back to the GIL.
2. Audit, then declare `py::mod_gil_not_used()` — pyAMReX first, since it owns the global state, then ImpactX/WarpX.
3. Stop shipping `cp313t`/`cp314t` wheels until the audit is done (they do work today, so this is a regression for users who already have them).

## Context

Found while fixing a separate free-threaded wheel bug — CMake FindPython does not select a free-threaded interpreter unless asked, so `cp314t` wheels were being built against a GIL-enabled Python (#611, and the same fix in BLAST-ImpactX/impactx#1612, BLAST-WarpX/warpx#7166, openPMD/openPMD-api#1927). With that fixed the wheels are now correctly built for `3.14t` — which is what surfaced this second, more fundamental question.

ax3l:
Closes #612.

> [!NOTE]
> Stacked on #611 (`Python_FIND_ABI`) — review the second commit.
> Requires AMReX-Codes/amrex#5615; without it the second concurrent `MFIter` aborts the process.
> **Merge order:** #5615 first, then repoint AMReX, then this.

> [!WARNING]
> **Two lines must be reverted before this merges**, both marked `TEMPORARY` in
> commit `2e0bc7a`: `commit_amrex` in `dependencies.json`, and `pyAMReX_amrex_repo` in
> `cmake/dependencies/AMReX.cmake`. They point AMReX at the #5615 branch. Without them
> this PR does not compile in CI — `src/Base/ParmParse.cpp` calls `ParmParse::tableCopy()`,
> which only exists in #5615 — and before that, every concurrency test was skipped, so CI
> reported green while running exactly one of them. Pinned, the suite goes from
> *1 passed, 13 skipped* to **15 passed**. The repo URI has to move too, not just the SHA:
> FetchContent clones and then checks out, and a plain clone of `AMReX-Codes/amrex` cannot
> check out a commit that so far exists only on the fork. Once #5615 lands, revert both and
> set `commit_amrex` to the merged SHA.

## Summary

We publish `cp313t`/`cp314t` wheels, but the modules used the two-argument `PYBIND11_MODULE` form, so importing pyAMReX re-enabled the GIL process-wide:

```console
$ python3.14t -X warn_default_gil -c "import sys, amrex.space3d; print(sys._is_gil_enabled())"
RuntimeWarning: The global interpreter lock (GIL) has been enabled to load module
'amrex.space3d.amrex_3d_pybind', which has not declared that it can run safely
without the GIL.
True
```

The tag is a promise, not a flag, so most of this PR is the audit behind it — and the tests that keep it honest.

## The contract

Same rule NumPy follows: **pyAMReX's and AMReX's own global state is race-free; your data is yours to synchronize.**

Safe from any number of threads: working on distinct objects (including ones built from the same `BoxArray`/`DistributionMapping`), reading the same object, each thread driving **its own `MFIter`** even over the same `MultiFab`, creating and releasing zero-copy views, and querying `ParmParse`.

Yours to synchronize: writing the same data from two threads, exactly as with a NumPy array.

Main thread only: `initialize`/`finalize`; MPI-collective calls unless AMReX is built with `AMReX_MPI_THREAD_MULTIPLE=ON`; RNG reseeding (`init_random` forwards to `amrex::InitRandom`, which reseeds process-global state); `TinyProfiler`; external GPU streams; dynamic `MFIter` scheduling.

All of this is written up in the new `docs/source/usage/threading.rst`.

## Changes

- `py::mod_gil_not_used()` on all three dimensionalities.
- A mutex around `initialize`/`finalize`, so a contract violation serialises instead of corrupting the AMReX instance stack, `ParmParse.pretty_print_table()` renders to a string and hands that to `py::print()` instead of holding a `scoped_ostream_redirect` across the walk -- that redirect swaps `std::cout`'s stream buffer process-wide, so it could capture a concurrent `amrex::Print()`. `ParmParse.to_dict()` walks `ParmParse::tableCopy()`, a locked snapshot.
- `Programming Language :: Python :: Free Threading :: 2 - Beta` — deliberately not `3 - Stable` until this has soaked in CI across platforms.
- A second ctest target running the concurrency tests with `PYTHON_GIL=0`.
- `tests/test_free_threading.py` — 13 concurrency tests. Each asserts a computed value rather than the absence of a crash. Threads are spawned *inside* a test, never across tests, because the autouse `amrex_init` fixture wraps every test in `initialize`/`finalize`.
- `tests/test_free_threading_perf.py` — two scaling regression tests, off unless `PYAMREX_BENCH=1`.
- `tools/bench_free_threading.py` — scaling harness with a GIL-on/GIL-off compare mode: the same binary, one environment variable apart.

## Testing

CPython 3.14 free-threaded, built with `pyAMReX_amrex_src` against AMReX-Codes/amrex#5615.

| | |
|---|---|
| Import no longer re-enables the GIL | no warning, `sys._is_gil_enabled()` → `False` |
| Full suite, `PYTHON_GIL=0` and `=1` | 249 passed, 95 skipped — both |
| Concurrency tests, `--count=50` | 750 / 750 |
| CUDA (RTX A2000) | 15/15; 300/300 at `--count=20` |
| SYCL (Intel Iris Xe) | green |
| GIL-enabled CPython 3.14 build | 248 passed — tag inert, `cpython-314` ABI tag |
| ThreadSanitizer | no races in AMReX or pyAMReX code |

A Python-level per-box kernel over a 128³ domain in 64 boxes, 14-core / 20-thread CPU:

| threads | GIL enabled | GIL disabled |
|--:|--:|--:|
| 1 | 1.00× | 1.00× |
| 2 | 1.07× | 1.93× |
| 4 | 1.06× | 3.55× |
| 8 | 0.99× | 4.92× |
| 14 | 0.97× | 6.25× |
| 20 | 0.97× | **7.29×** |

A NumPy-vectorised kernel over the same data is memory-bandwidth-bound and scales weakly either way (2.6× at 20 threads without the GIL) — but *with* the GIL it degrades to 0.39× at 20 threads, the threads contending for the lock rather than computing.

The AMReX-side locking costs +0.2% single-threaded on an `MFIter` + `fill_boundary` loop, inside the ±1.6% run-to-run spread.

## Follow-ups

- Releasing the GIL around long AMReX calls is orthogonal (a no-op on free-threaded builds) — #613.
- **CMake floor.** In a conda-forge prefix carrying both `python3.14` and `python3.14t`, only CMake ≥ 4.1 finds `Development.Module` for the `t` ABI; 3.31.4 and 4.0.3 fail with *"Could NOT find Python (missing: Development.Module)"* while still locating the interpreter. `pyproject.toml` asks for `cmake>=3.24.0` — worth revisiting for the `cp3xxt` wheels.
- **dpnp has no `cp314t` build**, so `to_xp()`/`to_dpnp()` zero-copy views are unavailable on free-threaded SYCL until Intel ships one. CuPy does have one, so CUDA is unaffected.

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -56,6 +56,66 @@ jobs:
       run: |
         mpiexec -np 1 python3 -m pytest tests/
 
+  free_threading:
+    name: GNU@13 Free-Threaded Python
+    runs-on: ubuntu-24.04
+    env:
+      CC: gcc-13
+      CXX: g++-13
+    steps:
+    - uses: actions/checkout@v6
+    - uses: actions/setup-python@v5
+      name: Install free-threaded Python
+      with:
+        python-version: '3.14t'
+    - name: Dependencies
+      run: |
+        .github/workflows/dependencies/dependencies_gcc.sh 13
+        .github/workflows/dependencies/dependencies_ccache.sh
+    - name: Set Up Cache
+      uses: actions/cache@v6
+      with:
+        path: ~/.cache/ccache
+        key: ccache-${{ github.workflow }}-${{ github.job }}-git-${{ github.sha }}
+        restore-keys: |
+             ccache-${{ github.workflow }}-${{ github.job }}-git-
+    - name: Build & Install
+      run: |
+        export CCACHE_COMPRESS=1
+        export CCACHE_COMPRESSLEVEL=10
+        export CCACHE_MAXSIZE=800M
+        ccache -z
+
+        export CMAKE_BUILD_PARALLEL_LEVEL=4
+
+        python3 -c "import sys, sysconfig; assert sysconfig.get_config_var('Py_GIL_DISABLED'), 'not a free-threaded interpreter'; print(sys.version)"
+
+        python3 -m pip install -U pip
+        python3 -m pip install -U build packaging setuptools[core] wheel
+        # CMake >= 4.1 to find Development.Module for the free-threaded ABI:
+        # 3.31 and 4.0 locate the interpreter but not the 't' ABI's headers.
+        python3 -m pip install -U "cmake>=4.1" pytest pytest-repeat numpy
+        AMREX_MPI=OFF AMREX_SPACEDIM=3 python3 -m pip install -v .
+
+        ccache -s
+        du -hs ~/.cache/ccache
+
+    - name: The GIL stays off
+      run: |
+        python3 -X warn_default_gil -c "import sys, amrex.space3d as amr; assert sys._is_gil_enabled() is False, 'importing pyAMReX re-enabled the GIL'; print('GIL disabled:', not sys._is_gil_enabled())"
+
+    - name: Unit tests, GIL disabled
+      run: |
+        PYTHON_GIL=0 python3 -m pytest tests/
+
+    - name: Unit tests, GIL forced back on
+      run: |
+        PYTHON_GIL=1 python3 -m pytest tests/
+
+    - name: Concurrency tests, repeated
+      run: |
+        PYTHON_GIL=0 python3 -m pytest tests/test_free_threading.py --count=20
+
   gcc14:
     name: GNU@14 Debug
     runs-on: ubuntu-24.04
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -370,9 +370,15 @@ add_dependencies(pyAMReX_python_sources pyAMReX_${AMReX_SPACEDIM_LAST}d)
 # Tests #######################################################################
 #
 if(pyAMReX_BUILD_TESTING)
+    # test_free_threading.py is ignored here and gets its own test below: this
+    # one runs with OMP_NUM_THREADS=3, so on an AMReX_OMP=ON build its worker
+    # threads would each open a 3-thread OpenMP team -- the oversubscription
+    # docs/source/usage/threading.rst tells users to avoid, and the one
+    # configuration in which AMReX hands two host threads the same RNG.
     add_test(NAME pytest.AMReX
         COMMAND ${Python_EXECUTABLE} -m pytest -s -vvvv
             ${pyAMReX_SOURCE_DIR}/tests
+            --ignore=${pyAMReX_SOURCE_DIR}/tests/test_free_threading.py
         WORKING_DIRECTORY
             ${CMAKE_PYTHON_OUTPUT_DIRECTORY}
     )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -383,9 +383,11 @@ if(pyAMReX_BUILD_TESTING)
     # set PYTHONPATH and PATH (for .dll files)
     pyamrex_test_set_pythonpath(pytest.AMReX)
 
-    # Free-threading (PEP 703) concurrency tests, with the GIL forced off.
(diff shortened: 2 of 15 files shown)


ax3l:
Potentially alternative approach to explore: that the whole MFIter is threaded in AMReX is slightly coincidential, because in async compute models (when we compile for GPU), we already assume that the preparatory work inside the MFIter loop (accessing arrays, scratch pad handling, etc) is relatively little work compared to the ParalleFor and ParallelReduce ops we do inside it.

So we could explore if explictly launching threads closer to the core NumPy/etc. CPU compute operations would be an option, which do not call into AMReX at this point anymore. That means we would not need to make AMReX thread-safe on these prep interfaces https://github.com/AMReX-Codes/amrex/pull/5615 )

We could prototype this in C++ with std::thread s used around `amrex::For`/`amrex::ParallelFor` w/o OMP of the MFIter loop. Launch threads before the MFIter loop and use inside, close to the compute, but after the AMReX function preps of boxes and arrays.

ax3l:
Too big of a change for now to make all pybind-exposed AMReX APIs thread-safe. We'll do a more focused approach for now providing a pattern for threading for compute regions, and releasing the GIL in controlled manner where safe.

